PR #1, opened 2016-02-29
Fix move packets

StamBoom:
This will fix:

- Entity Relative Move
- Entity Look And Relative Move

http://wiki.vg/Pre-release_protocol#Entity_Relative_Move


Paulomart:
:+1:

PR #9, opened 2016-03-01
Added support for named sound effect packet.

HugoDaBosss:
Adds support for the named sound effect packet.
Also tweaked the finding of new packet ids a bit.


MylesIsCool:
Looks good, great work!

PR #18, opened 2016-03-01
Fix MalformedJsonException on opening renamed containers

StamBoom:
This will also fix GhostGuys issue

----------------------------------------------------------------------------------------------------------------
GhostGuy  Version: 0.3.3
Very nice plugin! But when i open a gui from another plugin i get this error
http://i.imgur.com/jYtpeZZ.png
-----------------------------------------------------------------------------------------------------------------

### DIFF
--- a/src/main/java/us/myles/ViaVersion/transformers/OutgoingTransformer.java
+++ b/src/main/java/us/myles/ViaVersion/transformers/OutgoingTransformer.java
@@ -263,6 +263,20 @@ public class OutgoingTransformer {
 
             return;
         }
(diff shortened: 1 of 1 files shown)


MylesIsCool:
Was this always Chat? By chance do old clients every send a string without quotes? (Do we need to convert this to JSON if no quotes & no { like we do with a lot of the other code?)

Thanks,

If I get time I'll try this otherwise as long as it's tested on renamed chests etc and works it'll be cool!

StamBoom:
The default chest window:
{"translate":"container.chest"}
But after renaming it:
"renamed"

PR #31, opened 2016-03-02
Fixed lots of errors if two players log in at the same time

SanderGielisse:
Netty should not store a players pipeline handler in a public field which is not specific for that connection, fixed.


MylesIsCool:
Thanks for finding this, I'm not going to pull it but instead I've implemented a similar solution. I feel that all needed information should probably be stored in ConnectionInfo and for now instead of having a remove function we just do it in the handler. (I don't really think we need to ever remove 1.9 from a 1.9 client.)

Thanks!

Myles

PR #33, opened 2016-03-02
Added that fancy message when 1.9 clients try to enter a Minecart or Boat

fillefilip8:
Here you go :)


StamBoom:
Minecarts does work. The only thing that does not work yet is leaving it (You'll stay inside the minecart client-sided)

fillefilip8:
I noticed that now :)

StamBoom:
:+1:

fillefilip8:
Well on the spigot page it says  "(Probably won't be fixed due to them being changed in 1.9, we'll add a fancy message soon!)"
So i added a fancy message :) Haha

fillefilip8:
Adding that message can be a placeholder until a real fix comes out (if you guys decide to make it work)

MylesIsCool:
If we don't have anything towards Minecarts / Boats being properly fixed by the end of the day I'll pull this change as we should notify them.

This won't be long term as hopefully we will get it done.


Thanks,

Myles

StamBoom:
#83

fillefilip8:
^

PR #50, opened 2016-03-02
Rewrite Metadata system to read the data from the bytebuf instead of

Paulomart:
using the last sent packet.


MylesIsCool:
Awesome work :) (Depending on bukkit less is definitely better!)

Issue #70, opened 2016-03-03
Kick Error


ChavitaHD:
Hello,

I never get this error, but a lot of my players that are beta testing my revamped server get this error, and then get kicked.

http://prntscr.com/aaj474

What does it mean, and how can I fix it?

HugoDaBosss:
Are you using any title/tablist plugin?
And what plugins are you using?

HugoDaBosss:
Do you have any tips for me to recreate this issue?

StamBoom:
Are You using bungeecord?

PR #77, opened 2016-03-03
Don't send malformed JSON anymore, preventing kicks.

HugoDaBosss:


HugoDaBosss:
This would prevent issues like this https://github.com/MylesIsCool/ViaVersion/issues/70

MylesIsCool:
Looks good, Thanks! (Better to validate here then explode haha

PR #87, opened 2016-03-03
Easier to update version and added a command.

fillefilip8:
Easier to update version and added a new API function "getVersion()".
And a command :O


StamBoom:
Maybe also for README.md?

src/main/java/us/myles/ViaVersion/api/ViaVersionAPI.java
```diff
@@ -3,6 +3,11 @@
...
 
 public interface ViaVersionAPI {
-
+    /**
+     * Is player using 1.9?
+     * @param player
+     * @return
+     */
     boolean isPorted(Player player);
+    String getVersion();
```
Paulomart: Missing whitespace

fillefilip8:
Added more yml info and remove author name from the message in the command.

Paulomart:
You should squash the commits

PR #88, opened 2016-03-03
Use never as default player collision

StamBoom:
Use never as default player collision

### DIFF
--- a/src/main/java/us/myles/ViaVersion/transformers/OutgoingTransformer.java
+++ b/src/main/java/us/myles/ViaVersion/transformers/OutgoingTransformer.java
@@ -428,6 +428,20 @@ public class OutgoingTransformer {
             output.writeBytes(input);
             return;
         }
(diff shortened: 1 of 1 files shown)


MylesIsCool:
Not sure about this change because anyone who isn't in a team will have collisions?

StamBoom:
Yeah but the problem now is, 1.8 can push 1.9 users and 1.9 users can't push 1.8 users

StamBoom:
But yeah, it's true. We probably should put everyone in a team while joining until some plugin changes it.

Paulomart:
I think this would be up to the server owner. We could add an option to add everyone on the server into one team..

Johandrex:
So why were this closed?

MylesIsCool:
It was decided that this patch would not be beneficial I believe, it'll be redone soon no doubt.

Johandrex:
Awesome, will donate then ;)

PR #102, opened 2016-03-03
Config.yml (Default file and messages)

HAlexTM:
With configversion you can reset config.yml, you can use other strings for messages etc.. (tell me if you want an enabler for Metrics)


MylesIsCool:
Rejecting due to package change for Utils.

PR #104, opened 2016-03-03
Config.yml (Default file and messages)

HAlexTM:
With configversion you can reset config.yml, you can use other strings for messages etc.. (tell me if you want an enabler for Metrics)


MylesIsCool:
We have a config now.

HAlex:
Hi, this is some weird pull request I did as a 14 years old, I'd love if a repository admin could delete this issue. Thank you.

PR #131, opened 2016-03-05
Fix book formatting and invalid book tag at empty book

StamBoom:
**Before:**
Before

**After:**
After

This will also fix invalid book tag on empty book


HugoDaBosss:
Looks nice! Will test it when I get home.

MylesIsCool:
Good job :)

PR #138, opened 2016-03-05
Fix commandblocks

StamBoom:
This will fix commandblocks. MinecartCommands are still broken.

**only 1.8 commandblock features work!**


HugoDaBosss:
Looks really good! Seems like a lot of work.

PR #139, opened 2016-03-05
Added proper update message (based on checkplys)

HugoDaBosss:


checkplys:
@HugoDaBosss Can I use some methods you used in my plugin "NeedsUpdate?"?

HugoDaBosss:
@checkplys don't see what's so special about this methods, but sure, as long as you credit me.

PR #154, opened 2016-03-06
Improve POM and fix dead code

sgdc3:


sgdc3:
Whoops forgot some stuff from my other project.

PR #159, opened 2016-03-06
Change package name from bossbarapi to lowercase.

HugoDaBosss:


StamBoom:
For some reason git made it like this, it was lowercase on my IDE

PR #162, opened 2016-03-07
General code improvements

phenomax:
Hey!
In this commit I generally improved several things.
So I switched the project to bytecode version 1.8 in order to use e.g. lambda expressions.
Further details can be found in the commit.

Keep on your work :)

Greez

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,13 +53,13 @@
                 </includes>
             </resource>
         </resources>
-		<!--
+        <!--
         <testResources>
             <testResource>
                 <directory>src/test/resources</directory>
             </testResource>
         </testResources>
-		-->
+        -->
 
         <plugins>
             <!-- Maven Java Compiler -->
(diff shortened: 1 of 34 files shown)


StamBoom:
This plugin will probably stay at java 7, because many hosts are still using it.

phenomax:
Acutally I am not sure how many are still using java 7. But nevertheless it would be a good idea to upgrade both, plugin and hosts :)

StamBoom:
@phenomax Yeah I agree but people will create issue reports if the plugin doesn't work because they are using some shared host that doesn't want to upgrade. I have no idea how many are still using 7. But there will probably be some.

StamBoom:
Don't get me wrong. I would love to use java 8 features, but I don't think it's a smart move at the moment.

phenomax:
yeah you are probably right. I ll lower my bytecode version and switch back to java 7.

MylesIsCool:
Really nice work, few naming mishaps here and there. Makes me realise there are some files that need removing >.> (woops)

phenomax:
Thanks. If you could need another helping hand: do not hestiate asking :)

phenomax:
@Z609 It is still there. I just removed the unnecessary entry in the BossBar interface.

PR #170, opened 2016-03-07
Fix portals working slow in survival 

HugoDaBosss:
For some reason git thinks I changed the whole ViaVersionPlugin.java, but I just registered a new listener.


rj0495:
Need to check if nether is enabled/Disabled in spigot.yml

MylesIsCool:
Not going to pull quite yet as I feel this fix feels to complex and there should be a simpler way of fixing as we must be doing something stupid.

phenomax:
+1 you should collapse your catch blocks and also asynchronize your iterating through all players because this heavily blocks the main thread

PR #171, opened 2016-03-07
Fix jukebox

Matsv:
I hoped I could blame @HugoDaBosss for breaking it. But sadly I couldn't


HugoDaBosss:
Looks good! 
And yeah I know I "broke" it, you could blame me, but it wasn't working before, so technically not breaking ;)

PR #175, opened 2016-03-08
Reimplement bossbar methods

Matsv:


phenomax:
Absolutely unnecessary commit/merge. The boss bar methods are still there and there is no sense just adding them to the interface ...

MylesIsCool:
The main API returns an instance of BossBar, it would seem counter intuitive not to have this in the interface especially.

phenomax:
I can't follow your thoughts. Without having specified this method in the interface, the api class has this method according to Lombok.

PR #192, opened 2016-03-11
Optimize packet searching

HugoDaBosss:


MylesIsCool:
to pull or not to pull that is the question

PR #200, opened 2016-03-12
Implement new chunk transformer

lenis0012:
This properly writes empty chunks and chunk unloads.
Fixes bugs where void worlds bug out when rendering chunks.

This bug was related to required chunks not being sent.


rj0495:
Does this fix my Signs problem?

lenis0012:
@rj0495 it might.
I don't know

PR #205, opened 2016-03-13
Fix commandblocks not able to change command on update

Matsv:
In 1.8, the server sent PLAY_UPDATE_BLOCK_ENTITY every time some commandblock get's updated. But if you open a commandblock window, it wouldn't get updated

In 1.9, it only updates when you open it. The commandblock window get updated every time PLAY_UPDATE_BLOCK_ENTITY is sent. This would mean, if some commandblock somewhere changed. It would change your window to that. Even if it was not the commandblock you opened.

Before: https://fileserver.matsv.nl/media?id=0ea0808544239fb7debf084c226ac132
After: https://fileserver.matsv.nl/media?id=12c66ee1bcc30b2e53a4d7155028feb3


HugoDaBosss:
Looks good!

PR #253, opened 2016-03-19
Add option to not log failures

NomNuggetNom:
There is some insane console spam, would be nice to get rid of it.


HugoDaBosss:
Idea is nice, but we have a debug mode for that. So if we would like to suppress all of them, we would use a global existent debug variable. But if we suppress all errors, we won't be able to find bugs, so we will keep this 'spam'. This error only occurs when people cancel SpawnEntity packets.

PR #260, opened 2016-03-22
Add config option to suppress 'Unable to get entity for ID' errors

HugoDaBosss:


HugoDaBosss:
I think we can leave the old design for the config like this in master, as this will soon be deprecated.

MylesIsCool:
Not going to pull to prevent sync issues.

PR #318, opened 2016-04-03
Nicer injectPacketHandler();

sinnlosername:
Nicer injectPacketHandler(); ^.^


HugoDaBosss:
Looks good, but please try to check your formatting. We recommend using IntelliJ, as it handles formatting better.

HugoDaBosss:
Oh and by the way, please check your settings when you commit because your author settings are wrong.

sinnlosername:
@HugoDaBosss 
I use IntelliJ but before commit I changed one thing using text editor ^^

I don't know where the author settings and what they are.

Matsv:
@sinnlosername https://help.github.com/articles/setting-your-username-in-git/

MylesIsCool:
I feel thing makes it less future proof, and I prefer not hard coding obfuscated variables due to how they can change

sinnlosername:
@MylesIsCool 
View at the latest commit :3

MylesIsCool:
@sinnlosername It looks good, my only suggestion would be to perhaps use the logger for the noMatch, or perhaps removing noMatch functionality and just including it in the null checks as it's not even used all the time so instead of:

"if (serverData == null) return;"
You have
"if (serverData == null){
   getLogger().warning("Unable to locate serverData, this version may be incompatible with ViaVersion");
   return;
}"

Otherwise it looks good so far

sinnlosername:
@MylesIsCool
Done.

src/main/java/us/myles/ViaVersion/ViaVersionPlugin.java
```diff
@@ -217,7 +237,10 @@ public synchronized void handleAdd(Object o) {
                         }
                     }
                 }
-            }
+
+            }.find(connection, connection.getClass(), ElementType.FIELD);
+
+            getLogger().info("Successfully injected!");
```
MylesIsCool: This doesn't really mean successfully injected so it shouldn't be here. It might not of injected properly due to late bind, this needs removing

sinnlosername: @MylesIsCool 
Ok. Is there no late-bind check? O.o

MylesIsCool: @sinnlosername It's done in the method down below all this, essentially we inject ListWrappers which handle when the connection is added to the list later on :)

sinnlosername:
@MylesIsCool 
Done. again. ^^

MylesIsCool:
It still bugs be how it's inconsitent across the errors, using nomatch and then not. Preferable it'd be cleaner if nomatch and logger was just removed and it was handled manually by the main plugin class
"```
if (connection == null) {
 -                getLogger().warning("We failed to find the ServerConnection? :( What server are you running?");		 +                getLogger().severe("We failed to find the ServerConnection? :( What server are you running?");
                  return;		                  return;
              }		              }
```" etc

**Comment by sinnlosername** [comment]
meep

**Comment by qlimax5000** [comment]
@sinnlosername
```
git config --global user.name sinnlosername
git config --global user.username sinnlosername
git config --global user.email sinnlosername@users.noreply.github.com
```

MylesIsCool:
As you requested I'm going to make a decision which is to decline this:

Why?

Although it's nice and looks good I can't guarantee it will have the same functionality as the others, as the current system has been designed with maximum compatibility.
Also I feel it adds new classes which really we're not planning to have a system like this anywhere else in the code.

Hope this is alright,

Myles

PR #320, opened 2016-04-03
No longer rely on the plugin name.

HugoDaBosss:


MylesIsCool:
Not going to pull for the following reasons:
It is designed this way to prevent reloads breaking certain aspects of it, as the plugin instance won't work across it but Bukkit getPlugin will. (Go try it ;))
Also punishes people who try to be a scumbag and change the name in plugin.yml

PR #345, opened 2016-04-10
Merging to use modules.

intronate67:


intronate67:
Supposed to be on fork, ignore this.

Issue #338, opened 2016-04-07
Autorespawn at death cause chunk to be unloaded for 1.9.x clients


xion87:
### What version of ViaVersion are you using? Type /ver viaversion
0.7.4
### What version of Spigot are you using? Type /ver
This server is running CraftBukkit version git-Spigot-e4d4710-e1ebe52 (MC: 1.8.8) (Implementing API version 1.8.8-R0.1-SNAPSHOT)>

### What plugins are you using? Type /plugins
AsyncWorldEditInjector, WorldEdit, HealthBar, Skript, Craftconomy3, ChatControl, SkQuery, 
BungeePerms, WorldGuard, AsyncWorldEdit, Nicky, DuckSuiteTeleports, ViaVersion, pvpstats, Vault, EnderMaxiChest, PvPManager, TitleManager, ChestShop, DuckSuiteHomes, ProtocolLib, AAC, DuckSuiteWarps, HolographicDisplays, LeaderHeads

### How does this error happen? login? Using an item? 
I am using PvPManager premium version, with a config that allow to avoid death screen and autorespawn at death, sometimes player will get unloaded chunk when they respawn and have to relog for fix it

### Is there an error in the console? Use pastebin.com. Is there a kick message?
No errors

gigosaurus:
This bug along with #292 appears to have been introduced by 0.7, as they did/does not happen in 0.6.8 or earlier.

MylesIsCool:
@gigosaurus Have you tested this with the versions you said? Are you using any plugins? Do you get the same issue as above?

I'm in belief that the changes to chunks between 0.6.8 & 0.7.0 have the same functionality and would not fix this issue.

gigosaurus:
@MylesIsCool I tested this with a server only containing the ViaVersion plugin and the following plugin:

```java
import org.bukkit.Bukkit;
import org.bukkit.Location;
import org.bukkit.craftbukkit.v1_8_R3.CraftServer;
import org.bukkit.craftbukkit.v1_8_R3.entity.CraftPlayer;
import org.bukkit.event.EventHandler;
import org.bukkit.event.Listener;
import org.bukkit.event.player.PlayerJoinEvent;
import org.bukkit.event.player.PlayerRespawnEvent;
import org.bukkit.plugin.java.JavaPlugin;

public class BreakPlugin extends JavaPlugin implements Listener {

	@Override
	public void onEnable() {
		Bukkit.getPluginManager().registerEvents(this, this);
[...]

PR #350, opened 2016-04-12
Don't unload chunks twice

Matsv:
You unload it here:
https://github.com/MylesIsCool/ViaVersion/blob/master/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/types/ChunkType.java#L131-L134

And over here:
https://github.com/MylesIsCool/ViaVersion/blob/master/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/packets/WorldPackets.java#L119-L125

That was not the case in 0.6.7

@gigosaurus Could you test if #338 still happens?


gigosaurus:
The commit in this PR doesn't appear to make any difference whatsoever.

Matsv:
Thank You

Issue #383, opened 2016-04-30
1.8.8: Spigot's anti-Xray and Orebfuscator do not obfuscate blocks for 1.9.x clients


catdisc6491:
### What version of ViaVersion are you using? Type /ver viaversion
0.7.9
### What version of Spigot are you using? Type /ver
Spigot 1.8.8-R0.1
### What plugins are you using? Type /plugins
FancyShop, KeepItems, TerrainControl, BarAPI, NoDupe, Buycraft, DisableCommands, SignExploitBlocker, Nicky, VoxelSniper, ViaVersion (duh), LightAPI, AsyncWorldEdit, Dynmap, Deadbolt, WorldEdit, Skript, ClearLag, ModifyWorld, WorldBorder, PermissionsEX, AAC, AntiAura, ConditionalCommands, McMMO, Vault, UltraCosmetics, AntiCommandTab, PerWorldInventory, PlotSquared, WorldGuard, BungeePortals, mcMMO-WorldGuard, ProtectionLib, VoxelSniperRegions, CheckMyInventor, Prism, MagicSpells, iTag, mcMMOAction, Multiverse-Core, VanishNoPacket, Multiverse-Portals, PopulationDensity, FurnitureLib, DiceFurniture, TagAPI
### Are you using any additional software like BungeeCord? If so, what software and version? (Please list your plugins as well)
Bungeecord 1.9 Snapshot:4872075:1145
### How does this error happen? login? Using an item? 
Neither Spigot's built-in anti Xray obfuscation nor Orebfuscator succeeded in obfuscating ores buried in walls, allowing 1.9 hacked clients to freely view all ores buried in walls by using hacked clients. Obfuscation works fine for 1.8 clients, but not for 1.9.
### Is there an error in the console? Use pastebin.com. Is there a kick message?
I'm not getting any error in the console on startup or shutdown from ViaVersion.

adamk523:
Any ideas why blocks aren't becoming obfuscated properly for 1.9 clients? I don't want to have to drop 1.9 support for my server, because word's starting to get out about the lack of anti-Xray for 1.9 and I've been going crazy monitoring everyone for days now...

catdisc6491:
Any ideas about why ores aren't getting obfuscated for 1.9 clients? I really don't want to have to drop 1.9 support for my server, but word's getting out about the lack of anti-Xray for 1.9 hacked clients and I've been going crazy the past few days trying to monitor everyone...

[...]

PR #393, opened 2016-05-10
Fix anti-xray for 1.9

Matsv:
#383


Matsv:
Woops will make reflection

PR #395, opened 2016-05-12
WIP: Fix Signs

### LINKED ISSUE #352: Signs sometimes don't render correctly with some ProtocolLib depending plugins
### What version of ViaVersion are you using? Type /ver viaversion
0.7.4
### What version of Spigot are you using? Type /ver
1.8.8
### What plugins are you using? Type /plugins
ViaVersion, ProtocolLib, Libsdisguises
### Are you using any additional software like BungeeCord? If so, what software and version? (Please list your plugins as well)
Nope
When logging in all signs appear to be blank when using ProtocolLib. The signs seem to work fine prior to the chunk rewrite.
Nope

Matsv:
A work in progress to fix #352 

It works a lot better now, but sometimes it just doesn't work yet. Suggestions and idea's are welcome how to make it better & make it work 100%

The problem:
In some cases the Sign packet get sent earlier then the chunk packet, that'd in 1.8 give 'Unable to locate sign at x,x,x'. Too bad that got removed when you need it. :disappointed:

### DIFF
--- a/src/main/java/us/myles/ViaVersion/ViaVersionPlugin.java
+++ b/src/main/java/us/myles/ViaVersion/ViaVersionPlugin.java
@@ -447,6 +447,11 @@ public class ViaVersionPlugin extends JavaPlugin implements ViaVersionAPI, ViaVe
         return getConfig().getBoolean("anti-xray-patch", true);
     }
 
+    @Override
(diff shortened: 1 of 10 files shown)


Matsv:
Closing for now.

PR #402, opened 2016-05-19
Add possibility to get the supported versions from the api and status response

Matsv:
Oh, and also make the maps/list concurrent inside ProtocolRegistry


MylesIsCool:
damn mats back at it again with these awesome updates!

PR #418, opened 2016-06-07
Make the main class cleaner by giving the config his own class

Matsv:


MylesIsCool:
Mats is now the janitor :O!

Issue #420, opened 2016-06-08
Potions are converted incorrectly from 1.8 to 1.9


joserobjr:
### What version of ViaVersion are you using? Type /ver viaversion
0.8.4

### What version of Spigot are you using? Type /ver
git-Spigot-5f38d38-12698ea (MC: 1.8.8) (Implementing API version 1.8.8-R0.1-SNAPSHOT)

### How does this error happen?
1.8 uses a byte flag based system to manage potion effects, it's not based on a fixed key or id, so the same potion effect can be obtained from many different data values.

For example you can get the same fire resistance potion on 1.8 with:
```
/give @p 1 3
/give @p 1 8227
/give @p 1 12323
/give @p 1 10275
/give @p 1 11299
```
And there are many more possibilities to get the exact same effect with different data values as long we don't change some specifics bits.

That's why it's incorrect to assume that only 8227 must be converted to `fire_resistance`.

This incorrect conversion affected my server because I use the Bukkit's potion API to create potions and it does not generate the same potion ids as the creative tabs.

This is an example of valid item stack on 1.8 that is converted incorrectly to 1.9, it's shown as a normal water bottle on 1.9.
```java
import org.bukkit.potion.*;
ItemStack brokenOn1_9 = new Potion(PotionType.SPEED).toItemStack(1);
```

I've already fixed it in a branch and I'm going to send a pull request soon :wink:

MylesIsCool:
We only port certain ids as they are the vanilla way of obtaining them. I mean if you have a better way to convert them that isn't ugly and gives same results, that's also cool. :)

PR #421, opened 2016-06-08
Fixes potion data conversion from 1.8 to 1.9

joserobjr:
As I said on https://github.com/MylesIsCool/ViaVersion/issues/420

I've made this convertor when spigot 1.9 was released. I adapted it for ViaVersion.

This change will still use the existing potion data registration when available, so it will not cause any impact where it's already working.

### DIFF
--- a/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/ItemRewriter.java
+++ b/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/ItemRewriter.java
@@ -248,13 +248,14 @@ public class ItemRewriter {
         if(cached != null) {
             return cached;
         }
+        if(damage == 0) {
+            return "water";
+        }
 
         int effect = damage & 0xF;
         int name = damage & 0x3F;
         boolean enhanced = (damage & 0x20) > 0;
         boolean extended = (damage & 0x40) > 0;
-        boolean drinkable = (damage & 0x2000) > 0;
-        //boolean splash = (damage & 0x4000) > 0;
 
         boolean canEnhance = true;
         boolean canExtend = true;
@@ -278,16 +279,12 @@ public class ItemRewriter {
 
 
             default:
+                canEnhance = false; canExtend = false;
                 switch (name) {
+                    case 0: id="mundane"; break;
                     case 16: id="awkward"; break;
                     case 32: id="thick"; break;
-                    default:
-                        if(drinkable) {
-                            id = "mundane";
-                        }
-                        else {
-                            id = "water";
-                        }
+                    default: id="empty";
                 }
         }
 
@@ -298,9 +295,6 @@ public class ItemRewriter {
             else if(canExtend && extended) {
                 id = "long_" + id;
             }
-
-            //if(splash)
-            //    potion.splash();
         }
 
         return id;
--- a/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/ItemRewriter.java
+++ b/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/ItemRewriter.java
@@ -213,11 +213,9 @@ public class ItemRewriter {
                     item.setId((short) 438); // splash id
                     item.setData((short) (item.getData() - 8192));
                 }
-                if (POTION_ID_TO_NAME.containsKey((int) item.getData())) {
-                    String name = POTION_ID_TO_NAME.get((int) item.getData());
-                    StringTag potion = new StringTag("Potion", "minecraft:" + name);
-                    tag.put(potion);
-                }
+                String name = potionNameFromDamage(item.getData());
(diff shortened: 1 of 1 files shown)


MylesIsCool:
This converts items from 1.8 id to 1.9 string, but does nothing to 1.9 string to 1.8 id. I'm unsure how the server works here because technically the item data id it sent to the client will not match the id that the client sends back (eg. moving an item).

Otherwise I feel the switch block needs to be formatted better with using spacing and new lines for the variables.

Also splash is commented out?

joserobjr:
> It does convert back, 1.9 uses simple string keys for potion IDs, a normal fire resistance potion is only possible with "fire_resistance" key on 1.9, it's not possible to detect which 1.8's data value created that 1.9 key because 1.9 ids and 1.8 data values are in an one to many relationship. But it makes no difference if the damage is 2 or 12323 because potions are not stackable and they produce the exact same effect. So it's fine to convert 1.9 to 1.8 using the same data value as creative tabs.

However I tested stack with many items and different damage values using Essentials plugin, it worked perfectly fine.

> Otherwise I feel the switch block needs to be formatted better with using spacing and new lines for the variables.
It's organized by one line per ID, that's why I skipped the case 7 and its line. It's very easy to read as it's now. Are you sure that you want to break readability with line breaks, I can do that in few seconds but will reduce the code maintainability...

Matsv:
I see you removed the line
```
 if (POTION_ID_TO_NAME.containsKey((int) item.getData())) {
```
Have you tested what will happen if you send it without any data?

joserobjr:
I removed it because potionNameFromDamage() will always resolve any data value to it's corresponding 1.9 id. No data means 1.8's `0` that means 1.9's `water`. 1.9 also removed the unused potions and replaced them by Uncraftable Potion. All that information is available on http://minecraft.gamepedia.com/Data_values#Potions , it's working fine on my server.

My tests was to:
1. Summon 3 stacks of stacked potions with different data values but with same effect
2. Attempt to stack them: It stacks for 100ms but it's immediately updated to the expected result, which is to replace the cursor with the stack from the clicked slot
3. Spawn potions from creative tabs: Works as expected
4. Spawn removed potions like Stinky Potion (name bits 63), Rank Potion (55), Sparkling Potion (47), Clear Potion (7), etc: They all resolves correctly to Uncraftable Potion (`empty`)
5. Create potions using Bukkit's Potion API (which is deprecated because of 1.9 features but works): Worked fine.

I did not find any issues yet, this change is being used all the time on my server, if you want to try the IP is sabotador.com, just let me know your nick and I set you on creative mode.

src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/ItemRewriter.java
```diff
@@ -245,10 +243,72 @@ public static void toClient(Item item) {
...
+                    case 32: id="thick"; break;
+                    default: id="empty";
+                }
+        }
+
+        if(effect > 0) {
+            if(canEnhance && enhanced) {
+                id = "strong_" + id;
+            }
+            else if(canExtend && extended) {
```
MylesIsCool: Formatting here, should be on same line as the if, otherwise makes it hard to read

src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/ItemRewriter.java
```diff
@@ -245,10 +243,72 @@ public static void toClient(Item item) {
...
+        int effect = damage & 0xF;
+        int name = damage & 0x3F;
+        boolean enhanced = (damage & 0x20) > 0;
+        boolean extended = (damage & 0x40) > 0;
+
+        boolean canEnhance = true;
+        boolean canExtend = true;
+
+        String id;
+        switch (effect) {
```
MylesIsCool: I'd prefer this formatted to the java standards,

http://www.oracle.com/technetwork/java/codeconventions-150003.pdf

src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/ItemRewriter.java
```diff
@@ -245,10 +243,72 @@ public static void toClient(Item item) {
...
+            case 10: id="slowness"; canEnhance = false; break;
+            case 11: id="leaping"; break;
+            case 12: id="harming"; canExtend=false; break;
+            case 13: id="water_breathing"; canEnhance = false; break;
+            case 14: id="invisibility"; canEnhance = false; break;
+
+
+            default:
+                canEnhance = false; canExtend = false;
+                switch (name) {
```
MylesIsCool: Same formatting here

MylesIsCool:
I've left a few comments in relation to the formatting, otherwise it seems like a good patch :)

Matsv:
Thank you for your pull request, I manually merged it to change the formatting for you. :smile:

PR #422, opened 2016-06-08
Update names from 'snapshot' to '1.10'

yzimroni:
Some classes and variable names are still "snapshot", so I've changed them to "1.10"


Matsv:
Thanks

PR #427, opened 2016-06-08
Correct hologram Y offset

Harry5573:
https://github.com/filoghost/HolographicDisplays/blob/master/HolographicDisplays/Plugin/com/gmail/filoghost/holographicdisplays/util/Offsets.java#L15 The offset is by default -1.25 for 1.8, however 1.9 clients require only -0.29. (-1.25 - -0.29) = -0.96 which should be the correct offset.


Matsv:
Are you sure about it that this is the case for small armorstands from the official minecraft code?

Harry5573:
@Matsv Yes

Matsv:
Thanks

PR #429, opened 2016-06-09
Use Myles new Magic transformer for the ArmorListener

Matsv:


MylesIsCool:
I remembered that I thought I missed something but could never find it, good job!

PR #443, opened 2016-06-22
Fix piston tile entity crash

Matsv:


MylesIsCool:
<3

PhanaticD:
still get crashed

Matsv:
This is a client sided bug sadly. You can use the build on the build server that provides a config option to rewrite the piston extension. We'll have to wait and see if mojang will patch it next version.

PhanaticD:
it still crashes with replace-pistons: true

Matsv:
Could you pastebin your crashlog?

PhanaticD:
http://pastebin.com/cVHfkv1b

looks like they were invisible piston heads to keep sand from falling when the hub was made in 1.7, i replaced them with barriers and no more crash

Matsv:
Great to hear that :)

ridalarry:
the issue still exist if there is a block with id 36 which I use to block the water from flowing on my hub server portals

TiTo-418:
@ridalarry Replace them with barrier blocks

Rida94:
@TiTo-418 I cant barriers will prevent the players from going in the portal

PR #447, opened 2016-06-23
Support 1.9.2 servers [WIP]

lenis0012:
This code basically reads 1.9 data so it can then be transformed to add
in the tile entities.
it geta a little funky when it already loaded it as a 1.8 chunk tho, so
i had to add some methods to the chunk api.

in short: I hate minecraft


lenis0012:
One of my later changes seems to have broken 1.9.2 support.
Tomorrow ill try to clean it up and fix that issue

lets call it a WIP for now

src/main/java/us/myles/ViaVersion/protocols/protocol1_9_3to1_9_1_2/Protocol1_9_3TO1_9_1_2.java
```diff
@@ -74,15 +78,19 @@ public void registerMap() {
                 handler(new PacketHandler() {
                     @Override
                     public void handle(PacketWrapper wrapper) throws Exception {
-                        ClientChunks clientChunks = wrapper.user().get(ClientChunks.class);
+                        ClientWorld clientWorld = wrapper.user().get(ClientWorld.class);
 
-                        ChunkType type = new ChunkType(clientChunks);
+                        Chunk1_9_1_2Type type = new Chunk1_9_1_2Type(clientWorld);
                         if (wrapper.isReadable(type, 0)) {
```
Matsv: Don't forget to remove this if, this will be always false on 1.9-1.9.2 servers.

PR #455, opened 2016-06-26
Possibility to force the json transformer

Matsv:
This will fix issues if the syntax is similar to json like {test}


src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/Protocol1_9TO1_8.java
```diff
@@ -52,8 +52,12 @@ public static String fixJson(String line) {
         try {
             gson.fromJson(line, JsonObject.class);
         } catch (Exception e) {
-            System.out.println("Invalid JSON String: \"" + line + "\" Please report this issue to the ViaVersion Github: " + e.getMessage());
-            return "{\"text\":\"\"}";
+            if (ViaVersion.getConfig().isForceJsonTransform()) {
+                return "{\"text\":\"" + line + "\"}";
```
MylesIsCool: This should still be constructed by gson using JsonObjects and a string primitive to avoid quote escaping issues?

PR #513, opened 2016-10-14
Fix wrong package name when shading libraries

Acquized:
In the jar `pom.xml`, the shadedPattern's Package Name was set to `viaversion`, but in every other module the Package name was `ViaVersion`. This resulted in a quite wierd Package hierarchy when decompiling the Plugin. This Pull Request fixes this.
Image

### DIFF
--- a/jar/pom.xml
+++ b/jar/pom.xml
@@ -47,15 +47,15 @@
                     <relocations>
                         <relocation>
                             <pattern>org.spacehq.opennbt</pattern>
-                            <shadedPattern>us.myles.viaversion.libs.opennbt</shadedPattern>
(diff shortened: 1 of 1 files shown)


Matsv:
This will break for people that currently use those libraries.

HeroCC:
Possibly consider this PR for the next major release? Assuming ViaVersion follows semver.

MylesIsCool:
@HeroCC I mean I'd personally like to get this fixed.

As far as I know I don't think any other libraries use ours but i'm not sure. It really depends on impact.


Does anyone else have thoughts on this? :)

Acquized:
Another way would be to rename every Package in every Module to `viaversion`, that would fix the Issue

PR #517, opened 2016-10-16
ViaProtocolSupport Fix

### LINKED ISSUE #449: ProtocolSupport is not compatible with ViaVersion & 1.9.4 server
### What version of ViaVersion are you using? Type /ver viaversion
0.9.4

### What version of Spigot are you using? Type /ver
https://gyazo.com/ac7b39613d13de8adfd4d169f4b85ed3

### What plugins are you using? Type /plugins
https://gyazo.com/9bc8ee0e825e39d960fbbc94a82c6ba9

Mc 1.9.4, protocolsupport and viaversion won't allow a 1.10 client to join the server

https://gyazo.com/9f30328c9cc146402ec4f3f7dff8733d

Maxqia:
\*Insert Evil Monologue Here\*, This unfortunately doesn't work with 1.9's ProtocolSupport though, but it works with 1.8 and 1.10's. This fixes #449.

### DIFF
--- a/bukkit/src/main/java/us/myles/ViaVersion/bukkit/classgenerator/ClassGenerator.java
+++ b/bukkit/src/main/java/us/myles/ViaVersion/bukkit/classgenerator/ClassGenerator.java
@@ -3,12 +3,16 @@ package us.myles.ViaVersion.bukkit.classgenerator;
 import javassist.*;
 import javassist.expr.ConstructorCall;
 import javassist.expr.ExprEditor;
+
+import java.lang.reflect.InvocationTargetException;
+
 import org.bukkit.Bukkit;
 import org.bukkit.plugin.Plugin;
 import us.myles.ViaVersion.api.ViaVersion;
 import us.myles.ViaVersion.bukkit.handlers.BukkitDecodeHandler;
 import us.myles.ViaVersion.bukkit.handlers.BukkitEncodeHandler;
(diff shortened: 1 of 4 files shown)


bukkit/src/main/java/us/myles/ViaVersion/bukkit/handlers/ChannelHandlerContextWrapper.java
```diff
@@ -0,0 +1,253 @@
...
+    @Override
+    public ChannelHandlerContext read() {
+        base.read();
+        return this;
+    }
+
+    @Override
+    public ChannelFuture write(Object o) {
+        if (o instanceof ByteBuf && isLatestVersion) {
+            buffer.writeBytes((ByteBuf) o);
```
MylesIsCool: Would this not create malformed packets?

(One glued to another) if it writes multiple?

Maxqia:
Updated with 1.9 support! (and your requested changes :P)

MylesIsCool:
@Maxqia Wow! Looks awesome, the only thing I would have to comment about is, is the package for ViaHandler / the name of it correct. Because we don't implement it on other platforms currently? Would it have a more suitable name and be better in just bukkit?

Just a thought :) otherwise really good job

Maxqia:
I was planning on making it also for Sponge, so I can use a native version of ViaVersion with ProtocolSupport, but that's for another day ;)

Issue #553, opened 2016-11-15
Inventory bug with 0 amount of item


tade159SK:
Hi, on our network we have a problem, that in inventory that is showing servers with players (Quantity of Item)
Item with amount o 0 (zero) is not showing in inventory in Client 1.11 and Server 1.8.8
We are using latest ViaVersion

tade159SK:
Screen from 1.11 Client: https://ctrlv.cz/shots/2016/11/15/WojX.png

MylesIsCool:
Could you test this with 1.11 (not really sure how)?

I suspect Mojang has removed items with 0 amount.

tade159SK:
Screen from 1.11 Client: https://ctrlv.cz/shots/2016/11/15/WojX.png
Screen from 1.8.9 Client: https://ctrlv.cz/shots/2016/11/15/ScNe.png

MylesIsCool:
Could you test this with 1.11 (vanilla server)?

This isn't a bug if mojang removed it.

tade159SK:
This can't be tested with vanilla server, because this Inventory is in plugin

MylesIsCool:
If you can't replicate it in vanilla, mojang probably removed it

MylesIsCool:
Closing as this honestly can't be fixed if mojang removed it.

PR #657, opened 2017-05-10
Fixing item count on 1.11, where size <= 0. (#553)

KernelFreeze:
This will change all items with negative or zero amount, so the 1.11 client can display them.


MylesIsCool:
Looks fine, thanks :)

PR #676, opened 2017-06-01
Improved the shield handling (1.9->1.8)

RaynLegends:
This should fix a problem where the shield appears while players use a bow (switching to a bow quicky while holding a sword), or at least it reduces that possibility.


MylesIsCool:
Looks fine, thanks for your help :)

PR #736, opened 2017-08-20
Update Most Dependencies

CDAGaming:


MylesIsCool:
The versions are dependent on compatible server platforms so it's not really worth updating as it breaks specific api usage. (In regards to certain guava methods / netty)

CDAGaming:
Well, I can Revert the Guava/Netty Dependency Version, but i'd mention this in the POM

MylesIsCool:
I don't really see any advantage in updating lombok nor javaassist as they work fine. (And there's no new features we would intend to use, it just means we'd have to test they still work)

PR #754, opened 2017-09-28
Improve inventory handling for Bukkit servers

### LINKED ISSUE #739: 1.12.x Client - Can't Push All Same Items on Shift + Double Click
Client: 1.12.1
Server: 1.8.8 - Spigot

Ex: When players try to push all cobblestone blocks to chest same time using shift + double click on another cobblestone block in inventory, can't achieve this.

mmxw11:
This improves inventory handling and fixes #739 for **bukkit** servers.
1.12+ clients behave differently with the WindowClick packet when used with quick move action. 
http://wiki.vg/index.php?title=Protocol&oldid=13223#Click_Window

Previously clients grab the item from the clicked slot **before** it has been moved however now they grab the slot item **after** it has been moved and send that in the packet. 
Because of this the slot will always be empty so the sent item is null which doesn't make any sense.  
< 1.12 servers require the client to send the moved item and then the server will compare it to see if it equals with the item that the server has (to prevent abuse or something?). 
This isn't the case on 1.12+ servers due they no longer require the item being sent in the packet. So in order to patch this we need to make the server think we sent the right item.

At first i didn't want to use scheduler but later on i figured that it'd be the best option because at the end the action will be executed on the main thread anyway. Also waiting a few ticks before sending it to the server we're able to send all the items at once from one runnable rather than creating multiple runnables.

### DIFF
--- a/bukkit/src/main/java/us/myles/ViaVersion/bukkit/protocol1_12to1_11_1/BukkitInvContainerUpdateTask.java
+++ b/bukkit/src/main/java/us/myles/ViaVersion/bukkit/protocol1_12to1_11_1/BukkitInvContainerUpdateTask.java
@@ -20,7 +20,7 @@ public class BukkitInvContainerUpdateTask implements Runnable {
     public BukkitInvContainerUpdateTask(BukkitInvContainerItemProvider provider, UUID uuid) {
         this.provider = provider;
         this.uuid = uuid;
-        this.items = Collections.synchronizedList(new ArrayList<>());
+        this.items = Collections.synchronizedList(new ArrayList<InvItemStorage>());
     }
 
     public void addItem(short windowId, short slotId, short anumber) {
--- a/bukkit/src/main/java/us/myles/ViaVersion/ViaVersionPlugin.java
+++ b/bukkit/src/main/java/us/myles/ViaVersion/ViaVersionPlugin.java
@@ -175,6 +175,11 @@ public class ViaVersionPlugin extends JavaPlugin implements ViaPlatform {
     }
 
     @Override
+    public TaskId runSync(Runnable runnable, Long ticks) {
+        return new BukkitTaskId(getServer().getScheduler().runTaskLater(this, runnable, ticks).getTaskId());
+    }
+
+    @Override
     public TaskId runRepeatingSync(Runnable runnable, Long ticks) {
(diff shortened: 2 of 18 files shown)


bukkit/src/main/java/us/myles/ViaVersion/bukkit/protocol1_12to1_11_1/BukkitInvContainerUpdateTask.java
```diff
@@ -0,0 +1,53 @@
...
+public class BukkitInvContainerUpdateTask implements Runnable {
+
+    private BukkitInvContainerItemProvider provider;
+    private final UUID uuid;
+    private final List<InvItemStorage> items;
+
+    public BukkitInvContainerUpdateTask(BukkitInvContainerItemProvider provider, UUID uuid) {
+        this.provider = provider;
+        this.uuid = uuid;
+        this.items = Collections.synchronizedList(new ArrayList<>());
```
mmxw11: Apparently Travis doesn't like this?

MylesIsCool:
I haven't had the time to test this out but the only thing I'd change from looking at the code would be making it a config option which defaults true. (See https://github.com/MylesIsCool/ViaVersion/commit/a64c21da8bd0aed181686116aa46c263b0a72fa6#diff-7301e6ecd14268d7b38adb8364c26fec for what needs doing and then also add to unsupported options https://github.com/MylesIsCool/ViaVersion/blob/master/sponge/src/main/java/us/myles/ViaVersion/sponge/platform/SpongeConfigAPI.java#L16)

I'm hoping to have some time within the next week to look at this, but if anyone else would like to I'd love to hear other peoples opinions :)

mmxw11:
Thanks for the suggestion. I'll implement it later this week when i have time.
Regarding that Travis build the only thing i can think of causing the compilation to fail is that diamond operator which is weird. Is this running under java 7/8 isn't it?

mmxw11:
Here you go added a config option. :)
Also the Travis build passed for some reason i couldn't use diamond operators(?).

MylesIsCool:
I've refactored some of the code to make it more understandable. I'm going to test it for a few days to ensure it runs smoothly otherwise it works fine :).

I've also cleaned up some of the task code / supported code (some of it wasn't needed as it throws a runtime exception).

PR #765, opened 2017-10-21
Backport fix for MC-111054

Pokechu22:
The max length of the locale was increased in 1.12-pre3; we need to undo that.

As part of the fix for [MC-111054](https://bugs.mojang.com/browse/MC-111054 "[MC-111054] Invalid lang in options.txt leads to connection error"), the max length of the locale was [raised to 16 (from 7)](http://wiki.vg/Protocol_History#1.12-pre3), and the client now makes sure that resource packs have names shorter than that length.  However, for older servers, it is still 7, and thus the server will reject it (and the client won't know that the pack's invalid beforehand).

The fix is to just silently lower the length.  The server doesn't actually use the locale anywhere, so it's fine to do this.  Those that do will only be likely to handle a max length of 7, so it's fine more or less.

**Note**: I haven't tested this code in-game at all.  But the pack attached to MC-111054 should be usable to test it.


Pokechu22:
I just realized I did this on the wrong version, one sec.

MylesIsCool:
Much appreciated, thanks :)

PR #770, opened 2017-11-09
Fix spam-clicks from 1.9+ players on 1.8 servers

joserobjr:
Spam-clicking will not lower our hands/items anymore. Also hides the attack indicator when attacking but it will still blink very quickly when you change the item in your hands.

### DIFF
--- a/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/packets/EntityPackets.java
+++ b/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/packets/EntityPackets.java
@@ -17,7 +17,9 @@ import us.myles.ViaVersion.protocols.protocol1_9to1_8.Protocol1_9TO1_8;
 import us.myles.ViaVersion.protocols.protocol1_9to1_8.metadata.MetadataRewriter;
 import us.myles.ViaVersion.protocols.protocol1_9to1_8.storage.EntityTracker;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.UUID;
 
 public class EntityPackets {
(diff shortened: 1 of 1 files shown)


mmxw11:
This is a nice change but is the entire array rewriting necessary?
Is it required that the attribute is the top element of the list.
I tried to add it to the end without rewriting the entire array and it worked just fine.

Logics4:
Meanwhile, if some of you wants to test this pull request, I've built a jar with it.
It's based on the dev branch of ViaVersion (until this commit: [Update to 17w45b](https://github.com/MylesIsCool/ViaVersion/commit/61836895b0507aeacfd84e0385fc9dec50118243), latest at the moment I'm typing this).
Keep in mind that Myles and Mats will most likely not support you if you have any bugs using this build.
[ViaVersion-1.4.0-17w45b-PR#770.zip](https://github.com/MylesIsCool/ViaVersion/files/1462843/ViaVersion-1.4.0-17w45b-PR.770.zip) (Click on this link to download.)
(GitHub doesn't accept .jar files directly so I had to zip it, extract the zip file and put the jar file in your plugins folder.)

MylesIsCool:
In regards to people with existing servers:
- Could this have negative impacts on servers that set the attack speed?
- Does this need to be sent for every entity?

x9nico:
Hi, I tested the version (Thanks @Logics4) and I confirmed of it is work.
1 - It will not adding impacts on servers 
2 - I don't think so

electronicboy:
* because it's set as the first in the list, any modifications afterwards should apply, BUT; really this kinda change should be behind a configuration option just in case, imho

* The most important part is that they're sent to the player itself, but by vanilla behavior they are sent to all players. This really has no effect on normal entities and would basically be ignored.

MylesIsCool:
Closing this for now, why?

- The structure of the code should be reworked to be clean (using data structures over Objects)
- This should also be a config option

PR #776, opened 2017-11-16
Attack Damage on Weapons instead of weired 1.9 stuff

Gerrygames:
In 1.8 Weapons looked like this: https://puu.sh/ynhhE/5f0c5d7f5e.png
In 1.9 Mojang added some weired formatted stuff: https://puu.sh/ynhgN/5f31bb5c15.png

This would be fixed for 1.9+ clients on 1.8 servers by this pull request ;)


Roadhog360:
Is this ever going to get merged?

Roadhog360:
@MylesIsCool Don't mean to be a bother, but this PR seems to work fine and is a good addition, and thus should be merged.

MylesIsCool:
It's possible we may merge it although I'm unsure if we want to modify the 1.8 protocol as it's considered pretty stable, the PR is missing a few things (that I probably should have noted here earlier) which I'm sure Gerry has learned a lot since he made this PR :P 

- Config option
- Cleanup (IntelliJ auto format is preferred to make it consistent with rest of the code)

MylesIsCool:
See above :)

PR #835, opened 2018-05-27
Chat message changes

mmxw11:
I added some options for how chat messages should be handled on older servers. 
These options are: (Default: TRUNCATE)
+# _SPLIT_
+# _TRUNCATE_ 
+# _PASS_THROUGH_
+# _CANCEL_

I had to edit the code how packets are sent to the server as i couldn't get this to work.                               https://github.com/MylesIsCool/ViaVersion/blob/875e5d725258a16b53180d5d609d82506d9a3d54/common/src/main/java/us/myles/ViaVersion/api/PacketWrapper.java#L480

**[BELOW IS TESTED ON BUKKIT SERVERS]**
I did some testing and as the javadoc says
`A Channel received a message. This will result in having the ChannelInboundHandler.channelRead(ChannelHandlerContext, Object) method called of the next ChannelInboundHandler contained in the ChannelPipeline of the Channel.`
I came to the conclusion that it skips the decoder and will call the next "ChannelInboundHandler" after the decoder which for instance is NetworkManager on bukkit servers.
I don't think this is the intended behavior or am i missing something?

Opinions are welcome.

### DIFF
--- a/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_11to1_10/Protocol1_11To1_10.java
+++ b/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_11to1_10/Protocol1_11To1_10.java
@@ -1,8 +1,12 @@
 package us.myles.ViaVersion.protocols.protocol1_11to1_10;
 
+import java.util.concurrent.TimeUnit;
+
 import com.github.steveice10.opennbt.tag.builtin.CompoundTag;
 import com.github.steveice10.opennbt.tag.builtin.StringTag;
 import com.google.common.base.Optional;
+
+import net.md_5.bungee.api.ChatColor;
 import us.myles.ViaVersion.api.PacketWrapper;
 import us.myles.ViaVersion.api.Via;
 import us.myles.ViaVersion.api.data.UserConnection;
@@ -16,10 +20,12 @@ import us.myles.ViaVersion.api.remapper.ValueTransformer;
 import us.myles.ViaVersion.api.type.Type;
 import us.myles.ViaVersion.api.type.types.version.Types1_9;
 import us.myles.ViaVersion.packets.State;
+import us.myles.ViaVersion.protocols.protocol1_11to1_10.chat.ChatMsgCharacterLimitAction;
 import us.myles.ViaVersion.protocols.protocol1_11to1_10.packets.InventoryPackets;
 import us.myles.ViaVersion.protocols.protocol1_11to1_10.storage.EntityTracker;
 import us.myles.ViaVersion.protocols.protocol1_9_1_2to1_9_3_4.types.Chunk1_9_3_4Type;
 import us.myles.ViaVersion.protocols.protocol1_9_3to1_9_1_2.storage.ClientWorld;
+import us.myles.ViaVersion.protocols.protocol1_9to1_8.Protocol1_9TO1_8;
 
 public class Protocol1_11To1_10 extends Protocol {
     private static final ValueTransformer<Float, Short> toOldByte = new ValueTransformer<Float, Short>(Type.UNSIGNED_BYTE) {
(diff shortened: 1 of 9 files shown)


MylesIsCool:
Hoping to get around to testing this PR in the next few days (just to keep you updated), just got to test the sending changes to ensure it doesn't break depending plugins.

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_11to1_10/Protocol1_11To1_10.java
```diff
@@ -361,10 +367,39 @@ public void registerMap() {
...
-                            wrapper.set(Type.STRING, 0, msg.substring(0, 100));
+                        if (msg.length() > ChatMsgCharacterLimitAction.CHARACTER_LIMIT) {
+                            ChatMsgCharacterLimitAction action = ChatMsgCharacterLimitAction
+                                    .getChatMsgCharacterLimitAction(Via.getConfig().get1_11ChatMsgCharacterLimitAction());
+                            final String[] lines = ChatMsgCharacterLimitAction.handleChatMessage(msg, action);
+                            if (lines == null) {
+                                wrapper.cancel();
+                                PacketWrapper wpacket = new PacketWrapper(0x0F, null, wrapper.user());
+                                //line breaks break the formatting use color codes before every other word
+                                Protocol1_9TO1_8.FIX_JSON.write(wpacket, ChatColor.translateAlternateColorCodes('&', "&cYour message &cwas cancelled &cbecause it &cexceeded the &ccharacter limit &c(" + msg.length() + "/" + ChatMsgCharacterLimitAction.CHARACTER_LIMIT + ")!"));
```
creeper123123321: Hardcoded message

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_11to1_10/chat/ChatMsgCharacterLimitAction.java
```diff
@@ -0,0 +1,55 @@
+package us.myles.ViaVersion.protocols.protocol1_11to1_10.chat;
+
+public enum ChatMsgCharacterLimitAction {
```
JarvisCraft: It might be better to introduce the related interface and implement it in this enum rather than put logic to switch-statement.

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_11to1_10/chat/ChatMsgCharacterLimitAction.java
```diff
@@ -0,0 +1,55 @@
...
+            String name = action.name();
+            if (actionType.equalsIgnoreCase(name)) {
+                return action;
+            }
+        }
+        return TRUNCATE;
+    }
+
+    public static String[] handleChatMessage(String msg, ChatMsgCharacterLimitAction action) {
+        switch (action) {
```
JarvisCraft: (see review-comment above) Might be better to call a specific method (abstract for enum or the one of its interface) instead of switching over variants.

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_11to1_10/chat/ChatMsgCharacterLimitAction.java
```diff
@@ -0,0 +1,55 @@
...
+            case SPLIT:
+                if (msg.startsWith("/")) { // this will not work with commands
```
JarvisCraft: String#indexOf(char) or String#isEmpty() + String#charAt(int) might go better here

MylesIsCool:
Going to close as this is rather stale :) Anyone is free to pick this us and contribute otherwise.

PR #844, opened 2018-06-14
Platform and Sponge changes

creeper123123321:


bukkit/src/main/java/us/myles/ViaVersion/bukkit/platform/BukkitViaLoader.java
```diff
@@ -95,4 +113,17 @@ public Item call() throws Exception {
         });
 
     }
+
+    @Override
+    public void unload() {
+        // todo restore providers
```
MylesIsCool: What does this mean :P?

(Otherwise this all looks good 👍 )

MylesIsCool:
Great stuff, thanks :)

PR #857, opened 2018-07-18
Skulls don't have 'Rot'; Painting names

PureGero:
Skulls on walls don't always have a 'Rot' tag if it's rotation is 0.
'DonkeyKong', 'SkullAndRoses' and 'BurningSkull' were displaying the default painting instead.


MylesIsCool:
Looks great, thanks for the help :D means a lot!

PR #892, opened 2018-07-20
Don't use Paper block placement patch in 1.12.

Logics4:
Apparently the bug that caused the block placement issues with Paper was fixed in 1.12 ([according to Aikar from its development team](https://vgy.me/r4Xyal.png)). So, with this ~~commit~~ PR the patch to fix it won't run if the server is running Paper 1.12 or higher (which means it will only be used in 1.11.2 and lower server versions).


MylesIsCool:
Cheers :)

Issue #893, opened 2018-07-20
Reddust color not translated for 1.13 clients


Gerrygames:
### What is the output link of /viaversion dump? 
https://dump.viaversion.com/ed58e1fa3c6fce9278a16265cdd8e0fe0db4b0ddd460eae01221711d498b127e

### Are you using any additional software like BungeeCord? If so, what software and version? (Please list your plugins as well)
BungeeCord and some custom plugins.

### How does this error happen? login? Using an item? 
When particles are spawned.

### Is there an error in the console? Use pastebin.com. Is there a kick message?
No.

### Description of the issue
Reddust particles in 1.13 are always colored randomly.

1.13:
![](https://puu.sh/AZp9l/3212915b27.png)

1.12.2:
![](https://puu.sh/AZpa3/95fb5dd778.png)

https://github.com/MylesIsCool/ViaVersion/blob/a0899d4776134f4ac0f44d03b6b6b8f2936e09cc/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_13to1_12_2/data/ParticleRewriter.java#L109
The particle color should not be randomized if the amount is set to 0 and the speed is set to 1. In this case RGB values should be set to the X/Y/Z offset values.

PR #902, opened 2018-07-21
Fix reddust particle color for 1.13 clients

Gerrygames:
Fixes https://github.com/MylesIsCool/ViaVersion/issues/893


Matsv:
Thanks!

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_13to1_12_2/packets/WorldPackets.java
```diff
@@ -242,6 +244,24 @@ else if (particleId == 36)
                             return;
                         }
 
+                        //Handle reddust particle color
+                        ifStatement:
```
MylesIsCool: Does the label need to be here?

Otherwise great stuff :)

Gerrygames: It's for the break statement right here:
https://github.com/MylesIsCool/ViaVersion/pull/902/commits/bb4b5d3080865a4cc31013f90ca2a20041e08300#diff-1013ef390334502300ca050e5e285756R252

PR #910, opened 2018-07-21
Fix bungee-chat italic, bold, obfuscated, underline, strikethrough

creeper123123321:


MylesIsCool:
Great stuff thanks :)

PR #923, opened 2018-07-22
Named sound effect rewriting + Effect rewriting

creeper123123321:


common/src/main/java/us/myles/ViaVersion/protocols/protocol1_13to1_12_2/data/NamedSoundRewriter.java
```diff
@@ -0,0 +1,152 @@
...
+        oldToNew.put("record.stal", "music_disc.stal");
+        oldToNew.put("record.strad", "music_disc.strad");
+        oldToNew.put("record.wait", "music_disc.wait");
+        oldToNew.put("record.ward", "music_disc.ward");
+
+    }
+
+    public static String getNewId(String old) {
+        String newId = oldToNew.get(old);
+        return newId != null ? newId : old;
```
JollyajaX: I would recommend doing `old.toLowerCase()`. I've noticed a few sounds in 1.13 have the same name, but with caps removed (for example `item.fireCharge.use` is now `item.firecharge.use`). Sending caps to the client will result in a disconnect.

MylesIsCool:
Thanks for all your work towards this :)

PR #962, opened 2018-08-08
add support for 18w32a

ForceUpdate1:
Added support for the new blocks
Kown issue: Item remap is missing
New Item Id Mappings at 18w32a

0-442 => 0-442

443+ => id + 5

Special Thanks 
@Gerrygames and @Pokechu22


MylesIsCool:
Great stuff, thanks :)

PR #1035, opened 2018-09-30
Remove unncessary doubled map check

KennyTV:
These double checks can usually be simplified by removing the explicit check, as the contains check is nothing but a get() != null check.
... If accepted and plausable, I could also go over the other double checks


MylesIsCool:
Good catch, thanks :)

PR #1036, opened 2018-09-30
 Remove some more unncessary/doubled map check

KennyTV:
Cleaned up the easy double checks, tested it on a lobbyserver (1.8 with 1.8-1.13 support) with about 50 players and looked over it a second and third time to make sure I didn't negate any if-clauses

### DIFF
--- a/bungee/src/main/java/us/myles/ViaVersion/bungee/service/ProtocolDetectorService.java
+++ b/bungee/src/main/java/us/myles/ViaVersion/bungee/service/ProtocolDetectorService.java
@@ -27,16 +27,19 @@ public class ProtocolDetectorService implements Runnable {
     public static Integer getProtocolId(String serverName) {
         // Step 1. Check Config
         Map<String, Integer> servers = ((BungeeConfigAPI) Via.getConfig()).getBungeeServerProtocols();
-        if (servers.containsKey(serverName)) {
-            return servers.get(serverName);
+        Integer protocol = servers.get(serverName);
+        if (protocol != null) {
+            return protocol;
         }
         // Step 2. Check Detected
-        if (detectedProtocolIds.containsKey(serverName)) {
-            return detectedProtocolIds.get(serverName);
+        Integer detectedProtocol = detectedProtocolIds.get(serverName);
+        if (detectedProtocol != null) {
+            return detectedProtocol;
         }
         // Step 3. Use Default
-        if (servers.containsKey("default")) {
-            return servers.get("default");
+        Integer defaultProtocol = servers.get("default");
+        if (defaultProtocol != null) {
+            return detectedProtocol;
         }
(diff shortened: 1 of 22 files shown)


MylesIsCool:
Looks good to me, need someone to double check over just in case I missed something :P

bungee/src/main/java/us/myles/ViaVersion/bungee/service/ProtocolDetectorService.java
```diff
@@ -27,16 +27,19 @@ public ProtocolDetectorService(BungeePlugin plugin) {
...
+        Integer detectedProtocol = detectedProtocolIds.get(serverName);
+        if (detectedProtocol != null) {
+            return detectedProtocol;
         }
         // Step 3. Use Default
-        if (servers.containsKey("default")) {
-            return servers.get("default");
+        Integer defaultProtocol = servers.get("default");
+        if (defaultProtocol != null) {
+            return detectedProtocol;
```
Plancke: This should probably return the default protocol?

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/storage/EntityTracker.java
```diff
@@ -59,13 +59,13 @@ public EntityTracker(UserConnection user) {
...
 
     public UUID getEntityUUID(int id) {
-        if (uuidMap.containsKey(id)) {
-            return uuidMap.get(id);
-        } else {
-            UUID uuid = UUID.randomUUID();
+        UUID uuid = uuidMap.get(id);
+        if (uuid == null) {
+            uuid = UUID.randomUUID();
             uuidMap.put(id, uuid);
```
Plancke: ComputeIfAbsent would probably be faster here.

bungee/src/main/java/us/myles/ViaVersion/bungee/service/ProtocolDetectorService.java
KennyTV: My bad, would have completely missed it.. Thanks!

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/storage/EntityTracker.java
KennyTV: The project is set to Java 7 afaik

Plancke: Oh yeah just checked and it is, that's unfortunate ^^

MylesIsCool:
Thanks @KennyTV @Plancke much appreciated <33

PR #1045, opened 2018-10-12
Velocity platform

creeper123123321:
- [x] Make injector work
- [x] Changing from different server version
- [x] Port patches
- [x] Test
- [x] Fix concurrency issues

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -17,7 +17,7 @@
             <groupId>org.yaml</groupId>
             <artifactId>snakeyaml</artifactId>
             <version>1.18</version>
-            <scope>provided</scope>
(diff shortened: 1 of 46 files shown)


common/pom.xml
```diff
@@ -17,7 +17,7 @@
             <groupId>org.yaml</groupId>
             <artifactId>snakeyaml</artifactId>
             <version>1.18</version>
-            <scope>provided</scope>
+            <!--<scope>provided</scope> Velocity -->
```
creeper123123321: Will not relocating snakeyaml cause problems?

MylesIsCool: I suspect it probably would if other plugins use different versions, it's not on Bungee / Bukkit as they provide a version that works.

PR #1050, opened 2018-10-21
Fix scoreboard colors

Plancke:
Fixes color changing mid line in scoreboards in 1.13 and above

Before:
image
After:
image


MylesIsCool:
<3 much love

TitanicFreak:
Just tried this on our network, MineSuperior, as we were suffering from the scoreboard issue. And it works beautifully. Here's the proof.

Before,
29140818355d9beed8c90fc3a0af82ef

After,
3610e9e5b8f3559bf31b80c3ce90e1f1

PR #1053, opened 2018-10-22
fix #1044

### LINKED ISSUE #1044: Kick with 1.13 client when editing a book
### What is the output link of /viaversion dump?
https://dump.viaversion.com/7d37d3955551d4af7c3a7227af6b28154178fa64254215fdda395ea30bf6dc21

Still trying to figure stuff out, but every time I edit a book with a 1.13 client on a 1.8 server (no error with 1.8) I have an empty pages tag in a protocollib NbtCompound wrapper and am directly kicked with the error below (with and without protocollib installed).
However, this only occurs on editing the book, there's no problem when directly signing it
Tested Viaversion 1.5.1 and .2, Windows and Linux, with and without BungeeCord as well as ProtocolLib

https://hastebin.com/atukilajuj.apache

Gerrygames:
This should fix the issue #1044.


MylesIsCool:
Ty :)

PR #1057, opened 2018-10-24
1.14

Gerrygames:
Basic 1.14/18w43b support. Joining works. Items, BlockStates and Entity Ids are remapped. ~BlockLight and SkyLight is not send, I guess the client is awaiting this information and is only rendering chunks after you mine/place a block.~ Lights are send in the new Light Update Packet ;)
Block Id and Biome Id mappings are still missing.


wesleyvanneck:
so this would work with protocolsupport ?

krusic22:
> so this would work with protocolsupport ?

It should.

PR #1075, opened 2018-11-08
18w45a

Gerrygames:
Things that are still missing:
-Remapping tamed ocelots to cats
-There are still some lighting issues :(
-~Sound mappings (the current ones are for 18w43c)~
-~Block id mappings (the current ones are for 18w43c; blockstate id mappings are correct)~

Nevertheless I would like the PR to be merged, so other people can contribute easier.


MylesIsCool:
Thanks :)

PR #1084, opened 2018-11-12
Faster palette

Gerrygames:
Improves performance of the palette in the ChunkSection (when setting blocks).

I did run a simple test to see the performance difference.
`ChunkSection section = new ChunkSection();

long time;
time = System.nanoTime();
for (int i = 0; i < 10000; i++) {
	section.setFlatBlock(0, 0, 0, i);
}

System.out.println(System.nanoTime() - time);

time = System.nanoTime();
for(int i = 0; i < 10000; i++) {
	section.setFlatBlock(0, 0, 0, i % 10);
}

System.out.println(System.nanoTime() - time);`

Old palette:
59922433 (= 59.9ms)
2106559 (= 2.1ms)

New palette:
5519727 (= 5.5ms, 90.7% increase)
969519 (= 1.0ms, 54.0% increase)


MylesIsCool:
Looks good to me, thanks :)

PR #1090, opened 2018-11-15
18w46a

Gerrygames:
Changed the protocol id and updated the block, blockstate and item mappings. Packets and sounds did not change.


MylesIsCool:
Ty Gerry :)

PR #1118, opened 2018-12-12
18w50a

Gerrygames:
There are still some issues that need to be resolved:
-Chunks are not loading correctly when joining the server (chunks do load when teleported somewhere else)
-I think there were some entity metadata changes

The pull request can be merged, so other people can work on fixing the issues.


MylesIsCool:
Thanks Gerry :)

PR #1142, opened 2019-01-08
Use ProtocolInfo#getServerProtocolVersion for block connection, use b…

creeper123123321:
…ungee and velocity version if not ported


MylesIsCool:
Thanks for fixing these :)

PR #1146, opened 2019-01-09
change attack speed to 1024

creeper123123321:
the faster animation is caused by optifine/forge, it doesn't happen in vanilla


electronicboy:
Given that optifine/forge are pretty common, is this really a necessity/does it actually have a real-world impact?

PR #1148, opened 2019-01-09
Add fixes for custom flowerstems and snowcollision for 1.13 clients

KennyTV:
Added two (false by default) config fields: The first option fixes the lack of collision of the 1-layer snow for 1.13 clients by replacing it as 2-layers for the client (before you could stand on a 1-layer snow even if there's no block below, which is used in custom builds as well as through some world generating failures in lower versions with flying snow, which means 1.13 clients will have a hard time walking above them).

The second option is just something "nice to have" and fixes custom decorative flower stems (used by many builders, at least in Germany).
Also changed the "1.9&1.10" lines in the config to "1.9+" and moved blockconnection to the matching config section

(Images for how the stems should actually look, else the top of the sunflower stems would be the actual sunflower top -> if enabled and there is a block above, it will always be a stem, without a block just the normal flower top)
ary9f
ta219


KennyTV:
going to make the snow thingy a little better

MylesIsCool:
Thanks for re-ordering the options too :), looks great thank you :D

PR #1152, opened 2019-01-14
[WIP ]Remove unnecessary blockconnection caching, improvements in startup

KennyTV:
Removed all ids that are 1.13 only in the blockData.json, which obviously will never be called upon (checked with 1.12 Via mappings internally if there is a mapping present and autoremoved if not).
The entries are used to check if a block is occluding or not and as there won't be 1.13-only blocks on servers under that version, they aren't needed
same for blockConnections.json


KennyTV:
messed up merging, reopening later

PR #1160, opened 2019-01-17
fix fence connections to stairs

Gerrygames:
I also removed waterlogged blocks from the ~~stair~~ fence data map.


MylesIsCool:
Ty :)

PR #1162, opened 2019-01-18
Fix bungee and velocity boss bar, implement Velocity channel workaround

creeper123123321:


MylesIsCool:
Assuming this works fine, probably need to remove the commented out code :)

MylesIsCool:
Ty!

PR #1167, opened 2019-01-19
Reduce memory use for blockconnections

KennyTV:
Removed all entries in blockData and blockConnections files that will in no case be called upon, as they don't have corresponding 1.12 mappings.

Also avoid going through every single 1.13 block mapping in every blockconnection provider while creating tens of thousands of wrapper objects on first init -> go through all connectionhandlers for each block key


common/src/main/java/us/myles/ViaVersion/protocols/protocol1_13to1_12_2/blockconnections/ConnectionData.java
```diff
@@ -16,195 +16,204 @@
...
+                blockConnectionData.put(id, blockData);
+            }
+        }
+
+        JsonObject blockData = MappingData.loadData("blockData.json");
+        JsonArray occluding = blockData.getAsJsonArray("occluding");
+        for (JsonElement jsonElement : occluding) {
+            occludingStates.add(keyToId.get(jsonElement.getAsString()));
+        }
+
```
KennyTV: Changes (other than formatting) in this file are from this line to 184 and the interface the bottom

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_13to1_12_2/blockconnections/BasicFenceConnectionHandler.java
```diff
@@ -1,17 +1,22 @@
...
 public class BasicFenceConnectionHandler extends AbstractFenceConnectionHandler {
 
-    static void init() {
-        new BasicFenceConnectionHandler("fenceConnections", "minecraft:oak_fence");
-        new BasicFenceConnectionHandler("fenceConnections", "minecraft:birch_fence");
-        new BasicFenceConnectionHandler("fenceConnections", "minecraft:jungle_fence");
-        new BasicFenceConnectionHandler("fenceConnections", "minecraft:dark_oak_fence");
-        new BasicFenceConnectionHandler("fenceConnections", "minecraft:acacia_fence");
-        new BasicFenceConnectionHandler("fenceConnections", "minecraft:spruce_fence");
+    static List<ConnectionData.ConnectorInitAction> init() {
```
PSandro: What about using `Set` instead of `List` here? To me it looks like the elements don't have to be ordered, but maybe I'm wrong. :thinking:

KennyTV: https://stackoverflow.com/questions/10799417/performance-and-memory-allocation-comparison-between-list-and-set As far as I know when only iterating using a list is enough (i.e. set also has their hashmap in the background)

PR #1208, opened 2019-03-04
Add option to truncate edited books with more than 50 pages on 1.14

KennyTV:
When editing a book, the client can write up to 100 pages in 1.14 wheras it was 50 before.
Some anti-exploit plugins ban on that behalf (especially used on 1.8), so I figured an option to truncate that would be useful


MylesIsCool:
Thanks :)

PR #1220, opened 2019-03-20
Use JVM time instead of wall-clock time

isokissa3:
Pretty straightforward change, wall-clock time should not be used to measure duration's as it can lead to issues when the underlying systems time changes. For example, if the system time changes backwards few seconds this could potentially lead to kicking users from server for "Sending too many packets"

### DIFF
--- a/common/src/main/java/us/myles/ViaVersion/api/data/UserConnection.java
+++ b/common/src/main/java/us/myles/ViaVersion/api/data/UserConnection.java
@@ -18,6 +18,7 @@ import us.myles.ViaVersion.util.PipelineUtil;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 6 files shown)


MylesIsCool:
I'm going to close this PR, it's a really good point although after discussion with a few contributors the main points would be against changing it would be:

- nanoTime is potentially much more costly than currentTimeMillis - https://stackoverflow.com/questions/19052316/why-is-system-nanotime-way-slower-in-performance-than-system-currenttimemill

- nanoTime may not be consistent across threads (which we need it to be) - https://stackoverflow.com/a/9314616

isokissa3:
I got little bit curious after you linked those and after googling and reading the comments on those stackoverflow questions it looks like its fully safe and fast to use nanoTime() on Java 7 and above based on this answer: https://stackoverflow.com/questions/510462/is-system-nanotime-completely-useless/54566928#54566928

Also the question answer you linked saying its not consistent across diff threads has comments saying that this has been resolved in modern software!

PR #1230, opened 2019-03-24
read nbt in 1.14 chunk, pose metatype, don't change ping protocol if …

creeper123123321:
…different from provider


MylesIsCool:
Ty

PR #1302, opened 2019-04-27
Workaround for packet order issue in neighbour chunk updates

creeper123123321:
Not tested


creeper123123321:
Tested and seems to work

MylesIsCool:
Ty creeper :)

PR #1303, opened 2019-04-27
Some minor optimization

KennyTV:
essentially just some good old map lookups


MylesIsCool:
Thanks Kenny, :) some great optimisations which will probably mean a lot in the long run

PR #1309, opened 2019-04-29
Fix paperpatch

KennyTV:
welp I'm stupid 👀


MylesIsCool:
on the plus side we didn't see many bug reports from this :D

Issue #1312, opened 2019-05-01
ResourcePack 1.13.2 & ViaBackwards


Tomographix:
**Describe the bug, provide any errors**
No particular error.
"Internal Exception: java.io.IOException: An existing connection was forcibly closed by the remote host"
OR
"Disconnected"

Log from Minecraft Launcher
`[12:39:54] [Client thread/INFO]: Connecting to mc.fruitservers.net, 25565
[12:39:55] [Client thread/WARN]: Unknown custom packed identifier: minecraft:register
[12:39:55] [Client thread/WARN]: Unknown custom packed identifier: minecraft:register
[12:39:56] [Netty Client IO #8/ERROR]: Could not deserialize minecraft:nbt
[12:39:56] [Netty Client IO #8/ERROR]: Could not deserialize minecraft:nbt
[12:39:56] [Netty Client IO #8/ERROR]: Could not deserialize minecraft:nbt`

You can't seem to login to a 1.13.2 server with ForceResoucePacks installed with 1.14 client. Server is running Bungee with latest version of ViaVersion (dev) and ViaBackwards (dev)

**How can we reproduce it?**
1. Load Minecraft 1.14'
2. Login to mc.fruitservers.net
3. Go to the Prison server

**Expected behaviour**
Should be able to login to the server.

**ViaVersion Dump:**
https://dump.viaversion.com/bea84345b5440569b54a5a83a27365b99670864cf8d2bae7492d313142d8aea6

**Additional server info**
Bungee, ServerListPlus, ViaBackwards, ViaVersion, FruitDiscordLink

Tomographix:
`08:33:52 [INFO] [H1N9] disconnected with: EncoderException : us.myles.ViaVersion.exception.InformativeException: Please post this error to http://github.com/ViaVersion/ViaVersion/issues
{Type: Chunk, Source #1: us.myles.ViaVersion.protocols.protocol1_14to1_13_2.packets.WorldPackets$7 (Anonymous), Source #0: us.myles.ViaVersion.protocols.protocol1_14to1_13_2.packets.WorldPackets$7$1 (Anonymous), Data: [], Packet ID: 33}
Actual Error:  @ io.netty.handler.codec.MessageToMessageEncoder:107`

Double checked I was on the latest version of ForceResourcePacks.

MylesIsCool:
Should be fixed in latest build, ty @creeper123123321

PR #1321, opened 2019-05-08
Not tested workaround for #1312

creeper123123321:


creeper123123321:
Seems to work, player isn't kicked anymore

PR #1330, opened 2019-05-13
Reduce duplication in various classes

kashike:


common/src/main/java/us/myles/ViaVersion/api/remapper/PacketRemapper.java
```diff
@@ -32,7 +31,7 @@ public void map(Type type) {
      * @param newType The new type
      */
     public void map(Type oldType, Type newType) {
```
KennyTV: Now called recursively

common/src/main/java/us/myles/ViaVersion/api/remapper/PacketRemapper.java
```diff
@@ -44,7 +43,7 @@ public void map(Type oldType, Type newType) {
      * @param <T2>        The new return type.
      */
     public <T1, T2> void map(Type<T1> oldType, ValueTransformer<T1, T2> transformer) {
```
KennyTV: Same as above

MylesIsCool:
Looks good, thanks :)

PR #1344, opened 2019-05-17
Don't try to send register messages without channels

astei:
Bukkit interprets this as attempting to register the channel "" (the empty string), which fails on Minecraft 1.13 and above.


MylesIsCool:
Ty :)

PR #1345, opened 2019-05-19
Try to fix memory leaks

creeper123123321:
Not tested


xxDark:
This does not fix memory leaks. You increment ref count by 1, and then instantly decrement it in `finally` block.

creeper123123321:
It should fix a memory leak which may happen when the method that uses the ByteBuf object throws an exception

xxDark:
Than exception should be handled properly to avoid this

PR #1348, opened 2019-05-20
Fix suffocation damage for 1.14+ players in hitbox patch

KennyTV:


MylesIsCool:
Ty

Benedani:
Looks like the issue still exists, if I'm getting this right `playerToggleSneak` is called whenever you press/release shift regardless of whether or not the client actually makes the player stand up when there's enough space.
Does `isSneaking()` not work for detecting sneaking or?

KennyTV:
> Looks like the issue still exists, if I'm getting this right `playerToggleSneak` is called whenever you press/release shift regardless of whether or not the client actually makes the player stand up when there's enough space.
> Does `isSneaking()` not work for detecting sneaking or?

Please open an issue with info on what exactly doesn't work as intended and provide a `/viaver dump` output (just tested sneaking under blocks and sneak-jumping again on 1.8 and 1.13 servers with 1.13/1.14 clients)

PR #1367, opened 2019-06-14
fix 1.13 chunk reading, fix 1.9 chunk type not using global palette

### LINKED ISSUE #1364: Player gets disconnected on join
https://hastebin.com/xabojozuku.rb

**How can we reproduce it?**
1. Login with 1.13

**Expected behaviour**
The player just tried to join and got disconnected. I guess this shouldn't happen ;^)

**Screenshots**
Got no screenshots of the screen or smth. Only the stacktrace (hastebin above)

**ViaVersion Dump:**
https://dump.viaversion.com/3fa7bb1a6c8f5d5c384f92fb680493dec63c95655e33b47722b8ff827dfb5395

**Additional server info**
I use 2 proxies and they are connected with my own jedis plugin. Both proxies got viaversion installed and some plugins which are coded by myself.

Gerrygames:
Fixes #1364

I have a replay file containing (at least) one affected chunk packet. ViaVersion fails to read it without this fix: http://paste.cytooxien.de/ojuputosin.makefile
With this fix all chunks can be read perfectly.
I also added a check that will throw an exception if the expected size (based on the given bitsPerBlock value) does not match the size of the read array. If there are more issues with reading chunks, this will make our lives a little bit easier :)

I also implemented the global palette in the ChunkSectionType1_9. 
I did not remove the check for bitsPerBlock values lower than 4 in this type, because there were no problems with it. However it might have to be removed too.

The client (at least 1.14.2) seems to not like chunk sections with bitsPerBlock < 4, so I kept the lowest value for writing at 4. I honestly don't even know where these chunks are even coming from. Maybe the 1.13.x client accepts those chunks? If so, we could implement a ChunkSectionType1_14 and allow the 1.13 type to write with bitsPerBlock < 4. But I think this is not necessary.

### DIFF
--- a/common/src/main/java/us/myles/ViaVersion/api/type/types/version/ChunkSectionType1_13.java
+++ b/common/src/main/java/us/myles/ViaVersion/api/type/types/version/ChunkSectionType1_13.java
@@ -5,6 +5,7 @@ import us.myles.ViaVersion.api.minecraft.chunks.ChunkSection;
 import us.myles.ViaVersion.api.type.Type;
 
 public class ChunkSectionType1_13 extends Type<ChunkSection> {
+    private static final int GLOBAL_PALETTE = 14;
 
     public ChunkSectionType1_13() {
         super("Chunk Section Type", ChunkSection.class);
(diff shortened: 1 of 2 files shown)


Gerrygames:
Uploaded this to our server network 2 days ago. There were no complains about random kicks anymore. This only tests the ChunkSectionType1_13, because our servers are 1.13.2 and 1.14.2.
Someone needs to test this in a 1.8.x or 1.9.x server with 1.9.x-1.12.2 clients.

Gerrygames:
The global palette in ChunkSectionType1_9 is working too. I forced the use of the global palette by removing the check ``if (bitsPerBlock > 8)`` and all chunks are loading fine on a 1.8.8 server with a 1.12.2 client.

MylesIsCool:
Thanks Gerry, I suspect this bug has been the cause behind quite a few issues that we've been unable to work out :)

Issue #1370, opened 2019-06-19
Infested blocks are unbreakable.


Aurelien30000:
**Describe the bug, provide any errors**
A clear and concise description of what the bug is. Can you https://hastebin.com the error?
When I want to break a monster_egg block with a pickaxe, it doesn't work, the action is "cancelled".

**How can we reproduce it?**
1. Login with 1.13.x or 1.14.x version on a 1.12.2 server (perhaps 1.8.8-1.12.2, I don't really know)'
2. Take a pickaxe (iron, diamond)
3. Try to break a monster_egg block (which spawns silverfish)

**Expected behaviour**
The block should be broken.

**Screenshots**

**ViaVersion Dump:**
- Type /viaversion dump, and put the link here.
https://dump.viaversion.com/d4a7c1b503e0019f8268b46a13f41fdb5a7786afb6b0f31a906c9cce2966c22f

**Additional server info**
Do you use a proxy (eg. BungeeCord)? What software do you use and what plugins?
Paper 1.12.2-b1615

Thanks in advance for your answer :)

Aurélien

MylesIsCool:
Fixed in latest dev

Aurelien30000:
Thanks :)

PR #1375, opened 2019-06-22
Fix 1.13 clients unable to break infested blocks (#1370)

KennyTV:


MylesIsCool:
Would it be worth adding a config option for this with default true?

KennyTV:
I don't think so. Other than the potentially slower dig time there are no side effects, which should be better than having indestructable blocks in any case 👀

MylesIsCool:
Ty

PR #1400, opened 2019-07-15
1.14.4-pre6 support

KennyTV:
_don't try to look at the fact I accidentally bumped the version with the first pre-release_


MylesIsCool:
Ty

PR #1401, opened 2019-07-16
Fix 1.14 client lag on join

### LINKED ISSUE #1329: Client-side calculation of lighning causes a lot of lag on 1.14 clients
The client calculates the lightning itself if the server doesn't send the light data, causing lags in client thread

KennyTV:
Should fix #1329, might need some testing to ensure there are no lighting bugs when the chunk is updated


KennyTV:
Tested on Cytooxien, our servers, as well as some others with no new lighting issues - client lag is fixed™

MylesIsCool:
Thanks Kenny :)

PR #1414, opened 2019-07-28
Fix 0 light for non full blocks

### LINKED ISSUE #1280: Lighting bug - visual bug: stairs slabs beacons
When joining with 1.14 I've noticed several visual issues

**How can we reproduce it?**
1. Login with 1.14 client on a 1.13.2 server'
2. place stairs/slab/beacon
3. disconnect
4. join again

**Screenshots**
1.13.2 client on the left, 1.14 client on the right.
Screenshot_20190424_163403

**ViaVersion Dump:**
https://dump.viaversion.com/66e0a01d5ec7bd38d0a0ab426eefdba0b2e9a95ac9c942d9aec433fb074925d1

**Additional server info**
BungeeCord-Bootstrap:1.14-SNAPSHOT:4cef6d1:1405
ViaVersion v 2.0.1 build 318

KennyTV:
Fixes #1280 

The commits should probably be squashed, I kind of spammed those commits 👀

One known "bug": If one of those blocks is on a x/z chunk border with no sky- or blocklight in its own chunk, the 0 light will still be there, since we'd have to cache other chunks sky- and blocklight, but also all positions of slabs, stairs, walls, etc., which isn't really feasible


KennyTV:
Removed the cache option, since its situation would only ever occur on chunk borders with no block/skylight around them (see below), for which an unnecessary amount of caching would have to be done while also only working half of the time (because of chunk load order (unless we'd cache even more))
0tv29

MylesIsCool:
It would probably be nice to have a config option considering it adds extra work, otherwise given feedback looks like it's all good to go 🎉

PR #1419, opened 2019-08-04
fix explosion block desync, fixes #1365

### LINKED ISSUE #1365: Explosions desync the client (1.14.2 client on 1.13.2 server)
When there is an explosion (for example creepers or tnt) that destroys blocks, the client often will see some blocks as destroyed that are not destroyed. When trying to move into these blocks the client will be bugging.

**How can we reproduce it?**
1. Use 1.14.2 for the client, 1.13.2 for the server
2. Use a creeper to destroy some blocks
3. Try to jump around in the hole created by the creeper to see the desync (if you are bugging in an invisible block and click there you will often see the block reappear on the client)
4. If it did not work repeat at step 2

**Expected behaviour**
You see exactly those blocks destroyed that got destroyed by the explosion.

**ViaVersion Dump:**
https://dump.viaversion.com/2c1b064362b9fd3d58a96af4945e44620b5e440ad0a218e90acbdfc4afec768e

**Additional server info**
- BungeeCord
- Paper git-Paper-624

Gerrygames:


MylesIsCool:
ty

PR #1448, opened 2019-09-10
Fix plugin message pattern

KennyTV:
Some smart guy should prolly still verify this (tho it should be good: `+` = once or more, then there's a mandatory `:`, and another once or more for all lowercase alphanumerical thingies).
Before, the `?` allowed to emit the left identifier (which isn't possible in CB) and the`*` also marked a "0 or more"

Fixes https://github.com/ViaVersion/ViaBackwards/issues/138 and https://github.com/ViaVersion/ViaBackwards/issues/128


MylesIsCool:
The regex looks fine to me :)

creeper123123321:
> then there's a mandatory :
Namespace isn't mandatory, and empty namespace and/or path are parsed without error
> (which isn't possible in CB)
It's accepted by vanilla

PR #1449, opened 2019-09-10
Fix entity equipment packet for 1.9+ clients

KennyTV:
see the comments in the code - should be good, but I still sent it to someone to test it


JakubBien:
Now everything works fine, thx :D.

PR #1462, opened 2019-09-26
Fix some sound rewrites

KennyTV:
Also made a little prettier by increments the id variable instead of having a separate one, so that the actual new sound ids can be checked for (instead of having to account for the otherwise created difference)


MylesIsCool:
Ty

PR #1469, opened 2019-09-29
Fix 1.14+ walk animation

### LINKED ISSUE #1406: Move animation not working
Server version: 1.12.2
Client version: 1.14.4
Viaversion version: 2.1.3

Reproduce:
1. join 1.12.2 server with 1.14.4 version
2. look at player movement

Video:
https://youtu.be/HkrAwlTR8EI

lukas81298:
The player walk animation will not be shown to a player using Minecraft 1.14 or later if the player has their health set to Float.NaN. Several anti damage-indicator plugins replace the players health with NaN before the metadata is sent to the other clients.
This PR fixes #1406 by replacing NaN health with a normal value if it was not disabled in the config. This change will only affect 1.14+ clients.

### DIFF
--- a/bungee/src/main/java/us/myles/ViaVersion/bungee/platform/BungeeViaConfig.java
+++ b/bungee/src/main/java/us/myles/ViaVersion/bungee/platform/BungeeViaConfig.java
@@ -330,6 +330,6 @@ public class BungeeViaConfig extends Config implements ViaVersionConfig {
 
     @Override
     public boolean is1_14HealthNaNFix() {
-        return false;
+        return getBoolean("fix-1_14-health-nan", true);
     }
 }
--- a/bungee/src/main/java/us/myles/ViaVersion/bungee/platform/BungeeViaConfig.java
+++ b/bungee/src/main/java/us/myles/ViaVersion/bungee/platform/BungeeViaConfig.java
@@ -327,4 +327,9 @@ public class BungeeViaConfig extends Config implements ViaVersionConfig {
     public boolean isNonFullBlockLightFix() {
         return getBoolean("fix-non-full-blocklight", true);
     }
+
+    @Override
+    public boolean is1_14HealthNaNFix() {
+        return false;
+    }
 }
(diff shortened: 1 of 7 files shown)


common/src/main/resources/assets/viaversion/config.yml
```diff
@@ -177,4 +177,6 @@ force-json-transform: false
 # Minimize the cooldown animation in 1.8 servers
 minimize-cooldown: true
 # Left handed handling on 1.8 servers
-left-handed-handling: true
\ No newline at end of file
+left-handed-handling: true
+# Fixes walk animation not shown when health is set to Float.NaN
+fix-1_14-health-nan: true
```
KennyTV: This should go in the section above (`MULTIPLE VERSIONS OPTIONS`)

bungee/src/main/java/us/myles/ViaVersion/bungee/platform/BungeeViaConfig.java
```diff
@@ -327,4 +327,9 @@ public boolean is1_14HitboxFix() {
     public boolean isNonFullBlockLightFix() {
         return getBoolean("fix-non-full-blocklight", true);
     }
+
+    @Override
+    public boolean is1_14HealthNaNFix() {
+        return false;
```
KennyTV: Why is this false? (Same for the other 2)

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_14to1_13_2/MetadataRewriter.java
```diff
@@ -39,6 +39,12 @@ public static void handleMetadata(int entityId, Entity1_14Types.EntityType type,
                 if (metadata.getId() > 5) {
                     metadata.setId(metadata.getId() + 1);
                 }
+                if( metadata.getId() == 8 && type.isOrHasParent(Entity1_14Types.EntityType.LIVINGENTITY)) {
```
KennyTV: Add a whitespace after the two `if`s and remove the one before metadata, but else the pr is good to go :p

PR #1480, opened 2019-10-08
Fix scoreboard colors on 1.13

### LINKED ISSUE #1255: Bug scoreboard colours for 1.13.2 clients
I'm working on a 1.8 server and the colors in the scoreboards are bugged. This bug is only valid for 1.13.X clients

image

I have already set the configuration "fix-teams-colours" to true

My viaversion dump : https://dump.viaversion.com/e684964c00c9c9662792d5d80f4dadc706595b978370eebde6fdf97a66aade4e

Thanks

MrMicky-FR:
Fix scoreboard colors for some plugins using non colors formatting like `§r` for players names (same as #1050 but with other color/formatting codes).

This should fix #1255, but I can’t test because the plugins are not public.

I tested with QuickBoard (wich use `§r`) - https://github.com/ViaVersion/ViaVersion/issues/1255#issuecomment-486296053 - and it works

Tests with QuickBoard v4.3.0, using 1.13.2 client on 1.8.8 server:
_Before:_
Before

_After:_
After


MylesIsCool:
Looks great, thank you for contributing this change. We'll get this tested and should be good to go.

PR #1566, opened 2019-12-11
IS_NON_FULL_BLOCK_LIGHT_FIX as constant for performance

XjCyan1de:


astei:
Note that #1528 is a much more general solution to this performance problem.

MylesIsCool:
Closing as abstraction contains a more unified solution to all config values

PR #1576, opened 2019-12-12
Remove unneeded indirection

astei:
This is the first PR where I've pulled changes from the private ViaVersion fork we use at Mineteria that is strongly focused on performance.

This PR removes indirection in (de)serializing NBT, reading strings, and reading/writing varint-prefixed byte arrays. This is a slight micro-optimization, but it's a start to us contributing more actively to ViaVersion.


KennyTV:
Could you target the abstraction branch, since this breaks VB and VR? (the primitive use for example has already been implemented there)

astei:
It is sufficient simply to revert ba08c3ad and have that be a separate PR. Would that be ok?

MylesIsCool:
> It is sufficient simply to revert [ba08c3a](https://github.com/ViaVersion/ViaVersion/commit/ba08c3ad2fa0ce17dd75964f0b6a37db280bfb6f) and have that be a separate PR. Would that be ok?

Could you clarify further? Is this to avoid breaking VB / VR?

astei:
Yes, it avoids breaking ViaBackwards and ViaRewind.

MylesIsCool:
Sounds good 👍

astei:
@MylesIsCool I've reverted it, this should be ready to go.

MylesIsCool:
Awesome, thank you!

PR #1599, opened 2019-12-20
Fix Bungee injection on Java 12+

KennyTV:
_this is fine_

Thanks to the beautiful and only slightly fat CodeMatrix for helping me


Justsnoopy30:
Yay!

Xernium:
Checks out. Wish we didn’t need to hack the injection at all. Oh well, this works fine

MylesIsCool:
Thank you!

Issue #1473, opened 2019-09-30
Statistics not loading on 1.14.4


TheCompieter:
**Describe the bug, provide any errors**
when trying too open statistics on 1.14.4 it won't load

**How can we reproduce it?**
1. Login on 1.14.4'
2. Click on 'Statistics'
3. It wont load :/

**Expected behaviour**
it should load the stats of the player

**Screenshots**
https://i.imgur.com/8oxUM6p.png

**ViaVersion Dump:**
https://dump.viaversion.com/327dab8f9edf3eafa0434e767124774f6eb33c27e9a6e25bb72c00b73e5c5740

**Additional server info**
I dont use viaversion, viabackward,viarewind on my proxy anymore cause it was causing wierd bugs

soshimee:
That happens to me too, I would like to see it getting fixed.

PR #1631, opened 2020-01-15
Convert (general) 1.12.2->1.13 statistics

KennyTV:
Left out the item, block, and entity stats, since they'd require some _extra_ special treatment with considerably more caching (so I don't think it's actually worth doing so)


KennyTV:
Semi-fixes #1473

PR #1637, opened 2020-01-22
Add api version to prevent legacy data loading

KennyTV:


KennyTV:
Has greater implications/root problems I'd rather focus on in abstraction

PR #1643, opened 2020-01-28
Option to (not) remap infested blocks to their normal variants

### LINKED ISSUE #1642: Infested blocks are sent to 1.13+ as their non-infested variants.
On a 1.8 server, with a 1.13.x - 1.15.x client, Infested/Monster Egg blocks are sent to the player as normal stone / stone brick / cobblestone blocks. This seems to be related to [this issue](https://github.com/ViaVersion/ViaVersion/issues/1370) and subsequent fix.

**How can we reproduce it?**
1. Login on 1.13 or higher on a 1.8-1.12 server
2. Place an Infested Stone / Stone Monster Egg
3. The block is instantly transformed into a non infested variant (viewable in F3 with which block you're looking at)

**Expected behaviour**
The client would see the infested stone as opposed to normal stone.

**Screenshots**
In these screenshots, infested stone is textured differently than normal stone, which works fine on 1.12 and below, but on 1.13 and above, the client just sees a plain stone block.
Looking at infested stone on 1.12.2
Looking at infested stone on 1.15.2

**ViaVersion Dump:**
- Type /viaversion dump, and put the link here.
https://dump.viaversion.com/84a547e2ce9d06e20425051898adce368d4a293461ff6fc59c16f80ecc81f565

**Additional server info**
Do you use a proxy (eg. BungeeCord)? What software do you use and what plugins?
Simple test network proxy. Should be unrelated to issue.

KennyTV:
Fixes #1642

### DIFF
--- a/bukkit/src/main/java/us/myles/ViaVersion/bukkit/platform/BukkitViaConfig.java
+++ b/bukkit/src/main/java/us/myles/ViaVersion/bukkit/platform/BukkitViaConfig.java
@@ -246,6 +246,11 @@ public class BukkitViaConfig extends Config implements ViaVersionConfig {
     }
 
     @Override
+    public boolean isInfestedBlocksFix() {
+        return getBoolean("fix-infested-block-breaking", true);
+    }
+
+    @Override
     public int get1_13TabCompleteDelay() {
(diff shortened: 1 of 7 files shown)


MylesIsCool:
Ty

PR #1646, opened 2020-02-01
Add link to krusics "Blocked Version Generator"

Jo0001:


Jo0001:
It would be very useful to include the link to  the generator for blocked-protocols to the config.

PR #1654, opened 2020-02-10
Add vines blockconnection to prevent unwanted climbing

### LINKED ISSUE #1640: Vine bug 1.15 on 1.8 server
1.13 or 1.15 version introduced a feature that allows players to climb throw vines despite not being attached to walls. 1.8 clients until 1.12 requires vine attached to a wall in order to allow climbing.

This gives and extra advantage for 1.13+ players which can be dangerous (allowing players to reach areas unreachable, for example).

**How can we reproduce it?**
https://youtu.be/xJyILcy9R7I

(Video shows first on 1.8 server with 1.8 client, and then with 1.15 client)

**Expected behaviour**
Don't allow 1.13+ players to climb up without being attached. If not possible, maybe not showing vines for those versions?

**ViaVersion Dump:**
https://dump.viaversion.com/7160537c013075856ecf6540721a00cacf350f7be9a618cfb1448942a914a2be

KennyTV:
Works by only sending vines that are either:
a) attached to a block or
b) have another vine above them that is attached to a block
Fixes #1640

### DIFF
--- a/bukkit/src/main/java/us/myles/ViaVersion/bukkit/platform/BukkitViaConfig.java
+++ b/bukkit/src/main/java/us/myles/ViaVersion/bukkit/platform/BukkitViaConfig.java
@@ -241,6 +241,11 @@ public class BukkitViaConfig extends Config implements ViaVersionConfig {
     }
 
     @Override
+    public boolean isVineClimbFix() {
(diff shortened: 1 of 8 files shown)


MylesIsCool:
Ty

PR #1680, opened 2020-03-18
Fix login packet reading on 1.16 servers, fix uuid nbt tags

KennyTV:
Works™️, open for better solutions (if there are any) :>
Edit: For larger changes a new baseprotocol should be done, but I'd argue for this small of a thing, this check if fine


Justsnoopy30:
Thank you!

PR #1683, opened 2020-03-21
fix skull items

Gerrygames:


common/src/main/java/us/myles/ViaVersion/protocols/protocol1_16to1_15_2/packets/InventoryPackets.java
```diff
@@ -131,12 +137,32 @@ public void registerMap() {
 
     public static void toClient(Item item) {
         if (item == null) return;
+
+        if (item.getIdentifier() == 771 && item.getTag() != null) {
+            CompoundTag tag = item.getTag();
+            CompoundTag ownerTag = tag.get("SkullOwner");
+            if (ownerTag != null) {
```
MylesIsCool: Would it be worth checking if "Id" is present?

KennyTV: The server would throw errors for missing subtags, so only willfully bad nbt data per packet sending would have it missing

MylesIsCool:
Ty!

PR #1687, opened 2020-03-22
Load mapping data asynchronously  🚀

ForceUpdate1:
The changes offer the possibility to load the mapping data asynchronously to speed up the starting of the server.

The function can be activated or deactivated in the config. The function is deactivated by default.


common/src/main/java/us/myles/ViaVersion/ViaManager.java
```diff
@@ -47,8 +47,23 @@ public void init() {
...
+        //Register Protocols
+        ProtocolRegistry.registerProtocols();
+
+        //Load Mapping data
+        if(!platform.getConf().isLoadMappingsAsync()){
+            Via.getPlatform().getLogger().info("Loading mapping data ...");
+            ProtocolRegistry.loadMappings();
+        }else{
+            Via.getPlatform().getLogger().info("Load mapping data async ...");
+            Thread thread = new Thread(() -> {
```
creeper123123321: I think using CompletableFuture.wait and a runnable for each protocol would be more interesting

KennyTV: Possiblty, that would then also need proper "api" for ViaBackwards usage

ForceUpdate1: Do you think it makes a significant difference then, because it still blocks because it was waiting for the mappings to load

creeper123123321: You can use a Multi threaded executor for the CompletableFuture

Govindass:
> 
> 
> You can't even let users connect before mappings have been loaded, so that's pretty pointless and _will_ cause issues (unless you block new clients from connecting until loaded, which makes it pointless again)

maybe other plugins can take the time to load themselves faster? so mappings load at same time as other stuff, instead of one by one

KennyTV:
> maybe other plugins can take the time to load themselves faster? so mappings load at same time as other stuff, instead of one by one

See https://github.com/ViaVersion/ViaVersion/pull/1689 for a go at async loading with loading distribution as well as api for VB compatibility

PR #1689, opened 2020-03-23
Asynchronously load mapping data

KennyTV:
Depending on how much the CPU has to do during server startup as well, this will generally cut 2 to a bit over 5 seconds of startup time (loading was down from 11 to 4 seconds on a minimal setup Paper 1.15 server with VV and VB, and about a 3 seconds cut on our production servers with quite a limited amount of cores).

Using a cached threadpooplexecutor, mappings will be loaded asynchronously, until a user of a higher/lower client version logs on, in which case it will block and wait for the load (it does not block with clients running the server's version).

If ViaBackwards is present, Via will also cache relevant mappings, so that they will not be loaded twice - **still need ideas on how to enable the caching, currently just done by checking if VB is present** (it's fine, but maybe there's a better way).

A more minor change is a wrapper for the protocolstate and channel id instead of using a pair with boxed Integers, since those maps are called quite frequently, see the first commit https://github.com/ViaVersion/ViaVersion/commit/5d614f14d0c16561ca4d8e4a660c8cdb5e29293d

### DIFF
--- a/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_13to1_12_2/Protocol1_13To1_12_2.java
+++ b/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_13to1_12_2/Protocol1_13To1_12_2.java
@@ -42,14 +42,15 @@ import java.util.Map;
 
 public class Protocol1_13To1_12_2 extends Protocol {
 
-    public static final PacketHandler POS_TO_3_INT = new PacketHandler() {
-        @Override
-        public void handle(PacketWrapper wrapper) throws Exception {
-            Position position = wrapper.read(Type.POSITION);
-            wrapper.write(Type.INT, position.getX());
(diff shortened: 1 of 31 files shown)


common/src/main/java/us/myles/ViaVersion/protocols/protocol1_13to1_12_2/Protocol1_13To1_12_2.java
```diff
@@ -1151,6 +1147,14 @@ public void handle(PacketWrapper wrapper) throws Exception {
         registerIncoming(State.PLAY, 0x20, 0x2A);
     }
 
+    @Override
+    protected void loadMappingData() {
+        MappingData.init();
```
creeper123123321: Couldn't we create multiple completable futures for each mapping and join them?

PR #1733, opened 2020-04-16
Add git commit hash to dumps

KennyTV:
Writes an implementation version into the manifest, then also put into dumps (i.e. `Implementation-Version: git-ViaVersion-3.0.0-SNAPSHOT:d9291e6`)


MylesIsCool:
Ty

PR #1735, opened 2020-04-19
Add fire connection handler

KennyTV:
Not perfect (doesn't cover all quirks), but would blow a bit out of proportion otherwise for now


FormallyMyles:
Ty!

PR #1742, opened 2020-04-22
convert height maps, fix 1.16 chunk reading

Gerrygames:


common/src/main/java/us/myles/ViaVersion/protocols/protocol1_16to1_15_2/packets/WorldPackets.java
```diff
@@ -82,6 +84,14 @@ public void registerMap() {
                             blockEntity.put(skullOwnerTag);
                         }
                     }
+
+                    CompoundTag heightMaps = chunk.getHeightMap();
```
KennyTV, line 88: Needs to be moved up, this code isn't reached if the chunk doesn't have block entities

PR #1743, opened 2020-04-23
Uncache UserConnection on channel close, say farewell to lombok

KennyTV:


common/src/main/java/us/myles/ViaVersion/util/ConcurrentList.java
```diff
@@ -54,16 +58,19 @@ public void clear() {
...
+            try {
+                ConcurrentList<E> clist = (ConcurrentList<E>) super.clone();
+                clist.modCount = 0;
+                Field f = ArrayList.class.getDeclaredField("elementData");
+                f.setAccessible(true);
+                f.set(clist, Arrays.copyOf((Object[]) f.get(this), this.size()));
+
+                return clist;
+            } catch (ReflectiveOperationException e) {
+                throw new RuntimeException();
```
creeper123123321, line 72: e should be wrapped

PR #1769, opened 2020-05-09
Add runnable list to be executed before the full init

KennyTV:
Necessary for the proxies that don't have sync ticking, since the pre-init and full-init happen directly after one another without any option to intercept


FormallyMyles:
Looks good :)

PR #1796, opened 2020-06-06
Fix module versions

KVAC:
It is not possible to build a project


KennyTV:
The master branch is dormant for another few weeks until 1.16 releases, you should do any changes you want to do based off the abstraction branch (or download a build from the ci if you just want that)

PR #1799, opened 2020-06-08
Use FastUtil collections for frequently called code

KennyTV:


common/src/main/java/us/myles/ViaVersion/util/fastutil/CollectionUtil.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Utility class to possibly wrap FastUtil collections for faster access.
+ * These should only be used for high access and low/no change collections, since resizing FastUtil collections is expensive.
+ */
+public class CollectionUtil {
+    private static final boolean FAST_UTIL = checkForFastUtil();
+
+    private static boolean checkForFastUtil() {
+        try {
+            Class.forName("Int2IntMap");
```
Earthcomputer, line 18: Is this really meant to be looking in the default package?

KennyTV:
One thing to note is that this will not enable on CB/Spigot, since it for some reason relocated it - Paper on the other hand disables its relocation. Since Paper has become more used than Spigot, and so that other platforms or plugins may shade FU, it's checked by the default package.

xxDark:
>  so that other platforms or plugins may shade FU

Can't you do that by yourself?

KennyTV:
Sure, people would love to download a doubly shaded dependency weighing over 20mb 👀

KennyTV:
Well, moved to shading only a small bunch of FU classes to use the few ones more freely

xxDark:
I see that you are trying to predict exact size of maps and sets, so maybe also set it's loadFactor to 1?

KennyTV:
Sounds good!

PR #1833, opened 2020-06-24
Only print InformativeException cases in codec handlers

KennyTV:


common/src/main/java/us/myles/ViaVersion/api/data/UserConnection.java
```diff
@@ -307,7 +310,9 @@ public void transformOutgoing(ByteBuf buf, Function<Throwable, Exception> cancel
      * @param buf            ByteBuf with packet id and packet contents
      * @param cancelSupplier Function called with original CancelException for generating the Exception used when
      *                       packet is cancelled
-     * @throws Exception when transforming failed or this packet is cancelled
+     * @throws CancelException      if the packet should be cancelled
+     * @throws InformativeException if transforming failed or this packet is cancelled
```
creeper123123321, line 314: Is this correct? (InformativeException when cancelling)

PR #1927, opened 2020-07-23
Remove "API" from name in fabric.mod.json

creeper123123321:
It's kinda of a library, but it has implementations too. I don't know how to categorize it


Earthcomputer:
I would categorize ViaFabric as the "mod" (implementation) and ViaVersion as the "api" (library)

creeper123123321:
For me when i hear API it sounds like ViaFabric would have a lot of code and would implement everything.

creeper123123321:
The core is actually on ViaVersion

PR #2007, opened 2020-08-16
Close the open inventory before opening a new one on 1.16

connorhartley:
On older versions of the server you could open an inventory with one already open and it would close the old one and open the new one. On 1.16 this does not appear to be the case and attempting to do this would result in both inventories closing. This broke all the inventory menus we had on Mineteria and caused similar issues on another larger server. As a result this change tracks the inventory ID that has been open and removes it when it has closed. An attempt to open an inventory while one is already open will make sure the previous inventory is closed first. This solved problems on both servers.

### DIFF
--- a/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_16to1_15_2/packets/InventoryPackets.java
+++ b/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_16to1_15_2/packets/InventoryPackets.java
@@ -6,6 +6,7 @@ import com.github.steveice10.opennbt.tag.builtin.ListTag;
 import com.github.steveice10.opennbt.tag.builtin.LongTag;
 import com.github.steveice10.opennbt.tag.builtin.StringTag;
 import com.github.steveice10.opennbt.tag.builtin.Tag;
+import us.myles.ViaVersion.api.PacketWrapper;
 import us.myles.ViaVersion.api.Via;
 import us.myles.ViaVersion.api.minecraft.item.Item;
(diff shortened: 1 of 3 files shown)


common/src/main/java/us/myles/ViaVersion/protocols/protocol1_16to1_15_2/packets/InventoryPackets.java
```diff
@@ -28,22 +31,50 @@ public static void register(Protocol protocol) {
...
-                map(Type.COMPONENT);
+                map(Type.VAR_INT); // Window Id
+                map(Type.VAR_INT); // Window Type
+                map(Type.COMPONENT); // Window Title
 
                 handler(wrapper -> {
+                    InventoryTracker1_16 inventoryTracker = wrapper.user().get(InventoryTracker1_16.class);
+                    if (inventoryTracker.getInventory() != -1) {
+                        // Close open inventory before opening a new one.
+                        PacketWrapper closePacket = wrapper.create(0x13); // 1.16 inventory close
```
KennyTV, line 42: Use the ordinal of the packet enum from ClientboundPackets1_16 here

connorhartley:
> How does the client react if you send a close packet without an inv being open? (If it doesn't care, we don't have to care either, but otherwise you'd need to include places like teleportation or dying to reset the id as well; iirc it closes the inventory without explicitly being told there)

The client does not appear to care. I tried sending a close packet with a random ID and it did nothing, didn't even post an error or anything in the client logs.

PR #2031, opened 2020-08-22
Rework inventory workaround to cancel animation if in an open inventory for 1.16

RednedEpic:
Partially reverts #2007 and instead replaces it with a better solution. Rather than closing the current inventory as a new one is sent, this instead cancels any arm-swing that the client sends when swapping inventories (explained more extensively [here](https://github.com/ViaVersion/ViaVersion/pull/2007#issuecomment-678675622)).

Before:
ezgif-7-625588efa4d2
As you can see, every time I click the stone which is supposed to open a new inventory, it just re-opens the one that opens when I left-click the glowstone. May be a bit difficult to see, but this is more easily observed from one I sent of this happening on The Hive.

After:
ezgif-7-ee3df8b306f5

This has been tested extensively and no unintended side effects have been observed as a result of it.


assasin541:
@KennyTV I should be tested implemented. :)

KennyTV (approved):
Looks good, thank you

PR #2107, opened 2020-09-24
Fix falling dust ID in ParticleMappings

connorhartley:
The ID was incorrect causing players to crash when attempting to spawn redstone dust particles. This corrects it to look for the `falling_dust` ID not the `dust` ID, like it was intended to do.

Error caused: https://gist.github.com/connorhartley/57691cd199b57edff126e10d40827308


KennyTV (approved):
Thanks!

PR #2120, opened 2020-10-06
fixed 1.11.1 -> 1.12 show_achievement translation

RaphiMC:
This PR fixes the 1.11.1 -> 1.12 translation of show_achievement. Tested on 1.8, 1.9.4, 1.10.2, 1.11.2 server.


KennyTV (changes requested):
Put `JsonElement value = hoverEvent.get("value")` into a locar var to reuse (and asJsonObject/asJsonPrimitive from that).
The if check should be `value.isJsonObject()`

RaphiMC:
Changed it!

KennyTV (approved):

PR #2136, opened 2020-10-14
fix entity attributes length prefix

DaMatrix:
This fixes a bug where an entity attribute discarded for having an invalid identifier would not cause the resulting packet's length prefix to be updated, resulting in a packet that reported having more attributes that it actually did.

It also makes invalid entity attribute messages be hidden by the `suppress-conversion-warnings` config option.


KennyTV (changes requested):
Thank you, looks good - just one minor request; Please move the wrapper set after the loop (and only set if actualSize != size)

DaMatrix:
Done, anything else?

KennyTV (approved):
Thank you!

PR #2156, opened 2020-10-26
Update protocol versions

MrMicky-FR:
This just adds the missing protocol versions in ranges and removes 1.6.3, 1.7 and 1.7.1 because according to [wiki.vg](https://wiki.vg/Protocol_version_numbers) these versions haven't been released and also these versions are not in the releases list in the official Minecraft launcher:
versions

I also just removed the unecessary condition in VersionRange

### DIFF
--- a/common/src/main/java/us/myles/ViaVersion/api/protocol/ProtocolVersion.java
+++ b/common/src/main/java/us/myles/ViaVersion/api/protocol/ProtocolVersion.java
@@ -17,7 +17,7 @@ public class ProtocolVersion {
     private static final List<ProtocolVersion> versionList = new ArrayList<>();
 
     // Before the Netty rewrite
-    public static final ProtocolVersion v1_4_6 = register(51, "1.4.6");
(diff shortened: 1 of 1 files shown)


common/src/main/java/us/myles/ViaVersion/api/protocol/VersionRange.java
```diff
@@ -17,7 +17,7 @@
     public VersionRange(String baseVersion, int rangeFrom, int rangeTo) {
         Preconditions.checkNotNull(baseVersion);
         Preconditions.checkArgument(rangeFrom >= 0);
-        Preconditions.checkArgument(rangeTo > rangeFrom && rangeTo > 0);
```
Earthcomputer, line 20: Why?

KennyTV: That's correct, I have that locally commited as well

KennyTV:
Readded the non-existing version to not break API / previously expected behavior just yet.

PR #2162, opened 2020-11-01
Fix sound ID shift in 1.11

lennardk:
This possibly got broken in the rewrite in dc62394.
Fixes ender pearl throw and experience bottle throw/pickup sounds.


FormallyMyles:
Thank you!

Issue #2167, opened 2020-11-03
1.9-1.12 blocks and items are not changed to stone


tomcraft:
**Describe the bug, provide any errors**
Backend: 1.8.8 paper
Via*: 3.2.0 installed on Paper side only

When picking 1.9-1.12 blocks and items in creative inventory they are keeped as-is in the player inventory, so they are see as air by the server but the player still see them and it can be problematic with items like elytra because they can be equiped and used.

Whereas 1.13+ items are correctly changed to stone by Via.

**How can we reproduce it?**
Login with 1.16 client on 1.8 server/
Switch to creative
Pick elytra and netherite sword
You can equip elytra and fly.
The sword has been replaced to stone.

This is reproducible with any 1.9-1.12 block or item.

**Expected behaviour**
When picking an 1.9-1.12 block or item in creative inventory, it should be changed to stone.

PR #2180, opened 2020-11-08
prevent 1.12- clients to creative-pick server-side unknown block/item

tomcraft:
solve #2167 issue 

This is the first time i get into ViaVersion internals to contribute, i don't know if I did things like you want it to be.


KennyTV (changes requested):
Rename the `invalid` booleans to something like `newItem`, since 'invalid' isn't really the appropriate term here

tomcraft:
changes made

KennyTV (approved):
Thank you

PR #2192, opened 2020-11-13
Update ViaManager.java

Seanoo1:
Otherwise opinion for who can't upgrade server.


KennyTV:
The message is plainly wrong if you're not actually running a proxy server, and the recommendation of running Via on the backend has nothing to do with dropped legacy support

PR #2218, opened 2020-11-30
use ArrayDeque, less ram for PacketWrapper.readableObjects

creeper123123321:


tomcraft:
You could use a reverse Iterator to prevent the Collections.reverse operation.

https://stackoverflow.com/questions/18760832/how-can-i-iterate-a-arraydeque-in-reverse

tomcraft:
Oh no I'm tired

tomcraft:
Oh no wrong field. But you can just iterate with index based loop instead of reversing the whole List

common/src/main/java/us/myles/ViaVersion/api/PacketWrapper.java
```diff
@@ -488,7 +490,10 @@ public UserConnection user() {
      */
     public void resetReader() {
         // Move all packet values to the readable for next packet.
-        this.readableObjects.addAll(0, packetValues);
+        Collections.reverse(packetValues);
+        for (Pair<Type, Object> obj : this.packetValues) {
+            this.readableObjects.addFirst(obj);
+        }
         this.packetValues.clear();
```
tomcraft, line 497: for (int i = packetValues.size() - 1; i >= 0; i--) {
            this.readableObjects.addFirst(this.packetValues.get(i));
        }
        this.packetValues.clear();

KennyTV (approved):

PR #2282, opened 2021-01-08
Allow to pass JSON string in item lore on server versions < 1.14

CatCoderr:
Current implementation parses lore as simple text, escaping json string if it exists.


CatCoderr:
Just tested it on 1.16 client on 1.12.2 server. Works as expected.
image

KennyTV:
Sorry, but we don't add features to servers that don't exist in their native version.

PR #2291, opened 2021-01-17
Fixed encoding not properly working with other packet reading libraries

terminalsin:
Injecting a message to byte encoder beforehand would break outbound packets in ViaVersion. This solution is a good fix which won't impact anything.

### DIFF
--- a/bukkit/src/main/java/us/myles/ViaVersion/bukkit/handlers/BukkitEncodeHandler.java
+++ b/bukkit/src/main/java/us/myles/ViaVersion/bukkit/handlers/BukkitEncodeHandler.java
@@ -54,8 +54,10 @@ public class BukkitEncodeHandler extends MessageToByteEncoder implements ViaHand
                     throw (Error) e.getCause();
                 }
             }
+            
+        } else {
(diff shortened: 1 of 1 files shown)


FormallyMyles:
Do you have an example of a plugin which uses this fix? Just want to be 100% :) 

Thank you otherwise

terminalsin:
https://github.com/artemisac/Artemis-Packet-API Requires this to function.

terminalsin:
@FormallyMyles Still waiting on this to be merge 👀

FormallyMyles:
> @FormallyMyles Still waiting on this to be merge 👀

After looking into this further, is the clear() needed? AFAIK the ByteBuf is always a freshly allocated ByteBuf so I would doubt it makes any difference removing it, but I might have missed something.

It would be worth also copying the code to https://github.com/ViaVersion/ViaVersion/blob/master/sponge/src/main/java/us/myles/ViaVersion/sponge/handlers/SpongeEncodeHandler.java for consistency.

terminalsin:
@FormallyMyles The clear is indeed not *needed* in the case the pipeline is clean.

PR #2298, opened 2021-01-21
Fix - client leaves from server when dust particle occurs in 20w49a-21w03a

LemonCaramel:
If used a double value for redstone particles in 21w03a, it is correct that the code should work without any problems.
but, [same errors](https://github.com/ViaVersion/ViaVersion/issues/2249) occurred in 21w03a.

[It seems that mojang decided to use the float value instead of double.](https://bugs.mojang.com/browse/MC-207900)

image


KennyTV:
https://github.com/ViaVersion/ViaVersion/commit/e679a0c2b061f10f93c6ed3677f03c1f5bf018d2

PR #2305, opened 2021-01-28
Rename 1.16.4 Protocol

FlorianMichael:
Protocol appropriately renamed


KennyTV:
We won't rename them like that anymore since it's bound to backfire and look horrible/confusing with more than just 2 versions having the same protocol version (and already looks messy with even just too).

PR #2323, opened 2021-02-06
Fixing block placement for players with a lot of CPS

McPlayHD:
Players with a lot of CPS would have problems placing blocks if they can click faster than every 50ms because right now there is a check for double InteractEvents and with this pull request this should be fixed. I'm also upping the cache for block interactions to 1000 because some players can click really fast or are using AutoClicker and we don't want them to cause too much noise.


KennyTV:
Superceded by https://github.com/ViaVersion/ViaVersion/pull/2437

PR #2331, opened 2021-02-09
Avoid some crashers

MacacoLew:


KennyTV:
What exactly is that supposed to prevent? We don't do any transformation on buffers without readable bytes, so does Minecraft's decoder itself as well. So as far as I can tell, it only really saves calling the Minecraft decoder and a list add method call?
Empty bufs are also often used by libraries using Netty as a "last packet" to close connections after a successful send.

MacacoLew:
You don't want Spam in the ViaVersion Bukkit Decoder i think

KennyTV:
But why? It's very first line is to skip unreadable buffers as well.

MacacoLew:
Now with this patch you can be sure that  will block some exploits

KennyTV:
Sorry, but skipping empty buffers would give no benefit at all, and going as far as to just close the channel is definitely out of scope for ViaVersion. This belongs into its own, dedicated software, and with this implementation, it would also break expected behavior of other packet libraries.

PR #2372, opened 2021-02-27
Fix Bunge Chat relocation (ExceptionInInitializerError)

LemonCaramel:
It is a deficient part of Bunge Chat relocation. (5e227cb)

image


KennyTV (approved):
Thanks, for some reason I had problems with that locally, but goofed somewhere I think

PR #2384, opened 2021-03-04
[1.14 -> 1.13.2] Fix handling of unknown container ids

### LINKED ISSUE #2259: Can't open inventory for 1.14 player!
ViaVersion/ViaFabric#88:
Describe the bug, provide any errors
When opening some inventories in lower version than 1.16, we sometimes get an error:
[16:31:48] [Netty Client IO #0/WARN] (ViaVersion) Can't open inventory for 1.14 player! Type: Chest Size: 54

Is this happening specifically on ViaFabric?
I think this is due to ViaVersion but not entirely sure.

How can we reproduce it?
Use ViaFabric for 1.16, enable client-side version and use 1.9, join funcraft.net, try to open a menu by right-clicking any item.

Expected behaviour
It should still open the inventory, I guess.

RaphiMC:
Fixes #2259

### DIFF
--- a/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_14to1_13_2/packets/InventoryPackets.java
+++ b/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_14to1_13_2/packets/InventoryPackets.java
@@ -55,7 +55,7 @@ public class InventoryPackets {
                 handler(new PacketHandler() {
                     @Override
                     public void handle(PacketWrapper wrapper) throws Exception {
-                        Short windowsId = wrapper.read(Type.UNSIGNED_BYTE);
+                        Short windowId = wrapper.read(Type.UNSIGNED_BYTE);
                         String type = wrapper.read(Type.STRING);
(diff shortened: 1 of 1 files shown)


common/src/main/java/us/myles/ViaVersion/protocols/protocol1_14to1_13_2/packets/InventoryPackets.java
```diff
@@ -108,6 +104,12 @@ public void handle(PacketWrapper wrapper) throws Exception {
                                 case "minecraft:shulker_box":
                                     typeId = 19;
                                     break;
+                                case "minecraft:container":
+                                case "minecraft:chest":
+                                default:
+                                    if (slots > 0 && slots <= 54)
```
KennyTV, line 110: Add braces

KennyTV (changes requested):

RaphiMC (commented):
Added the braces around the if condition

KennyTV (approved):
Thank you

PR #2415, opened 2021-04-01
Fixes delayed shield blocking

R0bbyYT:
We have noticed that blocking through viaversion is delayed.
The reason for this is that you send an interaction to ViaVersion with the sword in the main hand,
which then puts a shield in the offhand and the client then
starts an interaction packet with the offhand blocking.

## no-delay-shield-blocking false
unknown

A better solution would be to trigger the blocking directly via the main hand but additionally place the shield visually in the offhand.
This change also fixes the issue that sometimes the shield stays in your inventory when you block for a very short time.

## no-delay-shield-blocking true
unknown_!

### DIFF
--- a/api/src/main/java/us/myles/ViaVersion/api/ViaVersionConfig.java
+++ b/api/src/main/java/us/myles/ViaVersion/api/ViaVersionConfig.java
@@ -77,6 +77,13 @@ public interface ViaVersionConfig {
     boolean isShieldBlocking();
 
     /**
+     * Whether the player can block with the shield without a delay.
+     *
+     * @return {@code true} if non delayed shield blocking is enabled.
+     */
+    boolean isNoDelayShieldBlocking();
+
+    /**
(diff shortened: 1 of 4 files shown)


KennyTV:
Are there any downsides of to this approach you know of? Otherwise you would be good to remove the config option and to always use it.

LabyStudio:
> Are there any downsides of to this approach you know of (other than clientside animation I guess)? Otherwise you would be good to remove the config option and to always use it.

The anti cheat from GommeHDnet (Intave) has issues when you disable "shield-blocking" or enable "no-delay-shield-blocking",
which is why it is still configurable. However, they are currently working on a solution.

LabyStudio:
Should we enable it by default or remove the option entirely?

KennyTV (approved):
Keeping the old behavior by default is good given anti-cheat concerns outside of Intave. Thanks for the pr!

Jpx3:
We would appreciate the ability to toggle this feature for players individually as well, preferably via API.

PR #2416, opened 2021-04-02
Use GitHub API instead of Spiget API for Update checking.

retrooper:
I think the GitHub API would be more reliable than the Spiget API.


KennyTV:
GitHub is down more often than Spiget as far as I am concerned, so there isn't much of a point in changing this. You also (somehow) used Bungee Chat, which we don't shade/provide on in the common module.

PR #2417, opened 2021-04-02
Implement showShieldWhenSwordInHand option

LabyStudio:
We are trying to improve pvp on 1.8 servers for players on the latest Minecraft version.
So I implemented another way to block without delay because the feature of #2415 causes problems with anti cheats.

The idea is that you have a shield in your second hand while you have a sword in your main hand.
The shield disappears as soon as you switch to another item.

The blocking interaction is on the shield and causes a client-side slowdown which is the expected behavior of the anti cheat.
I added an option called "show-shield-when-sword-in-hand" to toggle this feature.

demo


common/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/packets/InventoryPackets.java
```diff
@@ -145,8 +164,28 @@ public void registerMap() {
...
+                        for (short i = 0; i < stacks.length; i++) {
+                            Item stack = stacks[i];
+
+                            // Check if it is the inventory of the player
+                            if (showShieldWhenSwordInHand && inventoryId == 0) {
+                                // Store items in slots
+                                inventoryTracker.getSlotToItemIdMap().put(i, stack == null ? 0 : stack.getIdentifier());
+
+                                // Sync shield item in offhand with main hand
+                                entityTracker.syncShieldWithSword();
```
KennyTV, line 184: No need to call syncShieldWithSword with each single item transaction, move this outside of the for loop

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/storage/InventoryTracker.java
```diff
@@ -31,7 +38,99 @@ public String getInventory() {
...
+     *
+     * @param mode Inventory operation mode
+     * @param hoverSlot The slot number of the current mouse position
+     * @param button The button to use in the click
+     */
+    public void handleWindowClick(byte mode, short hoverSlot, byte button) {
+        EntityTracker1_9 entityTracker = getUser().get(EntityTracker1_9.class);
+
+        // Don't listen to the second hand slot
+        if(hoverSlot != 45) {
```
KennyTV, line 64: Early return with if (==45) {return;}

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/storage/EntityTracker1_9.java
```diff
@@ -97,6 +98,25 @@ public void setSecondHand(int entityID, Item item) {
...
+        // Get item in new selected slot
+        int inventorySlot = this.heldItemSlot + 36; // Hotbar slot index to inventory slot
+        Integer itemId = inventoryTracker.getItemIdInSlot((short) inventorySlot);
+        int itemIdentifier = itemId == null ? 0 : itemId;
+
+        boolean isSword = Protocol1_9To1_8.isSword(itemIdentifier);
+        Item shield = new Item(442, (byte) 1, (short) 0, null);
+
+        // Update shield in off hand depending if a sword is in the main hand
+        setSecondHand(isSword ? shield : null);
```
KennyTV, line 117: This should be short-circuited; if the itemId is null/0, set to null immediately, else also only create the shield Item instance if it actually is a sword

KennyTV (changes requested):
Isn't this missing incoming creative inv packets?

LabyStudio:
> Isn't this missing incoming creative inv packets?

Sorry, I didn't notice that. It should work now!

KennyTV:
Moving your mouse while placing a sword with right click doesn't seem to update the item cache correctly

KennyTV:
I can't tell you what exactly is wrong, but after some dragging and hotbar swapping I could desync it 👀

LabyStudio:
> I can't tell you what exactly is wrong, but after some dragging and hotbar swapping I could desync it 👀

Hm. I tried that for 10 minutes straight with multiple people.
I will look into it

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/storage/InventoryTracker.java
```diff
@@ -34,4 +41,143 @@ public String getInventory() {
...
+    public void resetInventory(short windowId) {
+        // Reset the cursor state of the inventory
+        if (inventory == null) {
+            this.itemIdInCursor = 0;
+            this.dragging = false;
+
+            // Remove window from cache (Except players window)
+            if (windowId != 0) {
+                this.windowItemCache.remove(windowId);
+                System.out.println("cleared window " + windowId + ", we have " + windowItemCache.size() + " remaining");
```
KennyTV, line 54: 👀

common/src/main/java/us/myles/ViaVersion/protocols/protocol1_9to1_8/storage/EntityTracker1_9.java
```diff
@@ -89,14 +91,40 @@ public void setSecondHand(int entityID, Item item) {
...
+    public Item getItemInSecondHand() {
+        return itemInSecondHand;
+    }
+
+    /**
+     * It will set a shield to the offhand if a sword is in the main hand.
+     * The item in the offhand will be cleared if there is no sword in the main hand.
+     * @param forceUpdate Ignore the previous state of the second hand and force sync the shield
+     */
+    public void syncShieldWithSword(boolean forceUpdate) {
```
KennyTV, line 111: The parameter is only ever set to false in its usages

KennyTV (changes requested):

KennyTV (approved):
*Should* be alright, I'm going to wait 1-2 days in case anyone else ~~actually willing to go through 1.8 protocol code~~ has something to say on this.

KennyTV:
Alright then, fingers crossed, thank you for the contribution! 🐱‍🏍

PR #2436, opened 2021-04-14
Clean up build scripts using new Gradle features

jpenilla:


KennyTV (approved):
Thenk, looks superb. the adventure bundle adventure-api exclusion might look confusing to someone, but 😎

PR #2437, opened 2021-04-14
Improve 1.9->1.8 block place translation (and fix some bugs)

RaphiMC:
This fixes a lot of issues regarding block placing on 1.8 servers.
First of all this fixes the limitation of 20cps that was enforced by the way block place translation worked before (Would close #2323).
This also fixes issues on paper servers for 1.9-1.12.2 clients (for platforms, that don't run on paper directly, like Bungee or ViaaaS).
The new code also mimics a 1.8 clients block place packet flow better.

Here is a video where I compare the 1.8 block place with 1.9 client (with my PR): https://streamable.com/xkjkuf
Here is a video on a paper server (with the paper patch removed manually for testing and my PR): https://streamable.com/bgcro1

I tested this with many test cases and didn't notice any abnormal behaviour / bugs.
Clients used for testing: Between 1.9 and 1.16.5


FormallyMyles:
When you tested were you able to see double interacts when the fix wasn't in place?

FormallyMyles (approved):

RaphiMC:
There were no double interactions without my fix because the old code relied on packet timings to fix those (cancel all block place packets within 50ms of the first one) which had the disdvantage of limitting max cps of players to 20cps. My fix doesn't rely on any timings, instead it uses the hand value the client sends. My fixes also fix some block place packets being sent when there shouldn't be one (which caused the paperspigot issue in the first place)

creeper123123321 (approved):

KennyTV (approved):
Thank you

PR #2438, opened 2021-04-14
Fix Java 16 compatibility on Bukkit/Sponge 1.8

MrMicky-FR:
This fixes the `java.lang.reflect.InaccessibleObjectException: Unable to make protected void java.lang.Object.finalize() throws java.lang.Throwable accessible: module java.base does not "opens java.lang" to unnamed module @4950b38` when running with Java 16 on Bukkit or Sponge 1.8. I also removed the unchecked and deprecation warnings in the ReflectionUtil class


creeper123123321 (approved):

KennyTV (approved):
Thank you

PR #2440, opened 2021-04-16
feat(bungee): use only unsafe field modifications

Proximyst:
Shev on the ViaVersion Discord server pointed out to me that there is no
need to set the modifier field with Unsafe instead of just setting the
field in itself with it. This removes our dependency on the
implementation details of `MethodHandles.Lookup#IMPL_LOOKUP`, and in the
same go lets us have a single call to define a field modification that
is classified as "forceful".

This is definitely going to be a controversial change, as this expands
our use of Unsafe considerably, while it is commonly wanted to reduce
it. Alternatively, we could implement a fall-back on all JRE versions to
use this behaviour and prefer the current behaviour (except perhaps for
JRE 16+?) when possible.

This has a notable side-effect of removing the illegal behaviour warning
on startup with Java 11.

---

This has been tested with the following Java distributions:
- Amazon Corretto OpenJDK 8.282.08.1
- Amazon Corretto OpenJDK 11.0.10.9.1
- GraalVM CE 21.0.0.2 OpenJDK 11.0.10+8 (also with BungeeCord)
- Amazon Corretto OpenJDK 16.0.0.36.1

... with Waterfall b1556 and Linux 5.10.28_1.


creeper123123321 (approved):

KennyTV (approved):
Should be good! If really needed, we can make a fallback implementation of the interface later. 🐱‍🏍

PR #2446, opened 2021-04-20
Modified ProtocolVersion to properly reflect the fact that Minecraft versions 1.16.5 and 1.16.4 use the same underlying network protocol ID.

akemin-dayo:
Modified ProtocolVersion to properly reflect the fact that Minecraft versions 1.16.5 and 1.16.4 use the same underlying network protocol ID.

… Not much else to say about this, it's a pretty simple modification ;P


KennyTV:
Thank you for the pr - we have done some larger changes inbetween all the 1.17 work, so all current development is done in the dev branch, where we have done this already https://github.com/ViaVersion/ViaVersion/blob/dev/api/src/main/java/us/myles/ViaVersion/api/protocol/ProtocolVersion.java#L78.

akemin-dayo:
Oh oops, didn't see that. Thank you for letting me know!

PR #2447, opened 2021-04-23
Update Dependencies

linsaftw:
Fix vulnerabilities reported by analytics.

Updated to latest non-breaking netty version, guava and snakeyaml.

Tested on Bungee & Spigot with no errors.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,7 +111,7 @@
         <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-all</artifactId>
-            <version>4.0.20.Final</version>
+            <version>5.0.0.Alpha2</version>
             <scope>provided</scope>
             <optional>true</optional>
         </dependency>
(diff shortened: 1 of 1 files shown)


creeper123123321 (changes requested):
Netty 5.0.0 isn't finished and updating these dependencies don't change anything because it's from platform

astei (changes requested):
This PR is simply fractally wrong.

For one thing, you depend on ancient Netty 5, prior to its revival in late 2018 (I wish they'd called it Netty 6 to make the differences clear, but too late). The versions used for Guava and SnakeYAML ancient for a reason: ViaVersion has to launch on old versions of Minecraft, thus we can't depend on modern versions of Netty, Guava, or SnakeYAML, lest we use a feature that only exists on newer versions of these libraries (Guava is probably the best example).

linsaftw:
Sorry, wrong netty version. I intended to use 4.x.x instead (5.x.x doesn't work). Thanks for the answer!

Issue #1519, opened 2019-11-13
Allocated RAM amount in /viaver dump


creeper123123321:
**Is your feature request related to a problem? Please describe.**
Yes. There may be some issues related to low RAM and block connections but we can't know if it is related to heap and issue managers may be confused with the "weird" issues

**Describe the solution you'd like**
Adding allocated and used heap to /viaver dump

**Describe alternatives you've considered**
No one

**Which versions is this for?**
All versions

FlorianMichael:
Resolved in https://github.com/ViaVersion/ViaVersion/pull/2451

PR #2451, opened 2021-04-25
Added Runtime-Information to dump (implements #1519)

PierreSchwang:
The ViaDump now includes the runtime information like max, total and free memory.

Example Dump: https://dump.viaversion.com/9b281a34098d3743b6e988b6237b3715fd9a471b24f250616e519df69b691b78

Currently the unit for the memory is bytes - maybe KB / MB should be used instead? I don't know what would be the smartest solution


Camotoy:
I would do MB, since that's generally regarded as the balance of readable and flexibility of modification.

PierreSchwang:
The generated dump for latest change: https://dump.viaversion.com/a54a66d0ab0d4a8c64446852e540f7ebc971d3ecfdf3fee78bfc56b713433d56

Relevant parts are:
```json
"runtimeInfo":{
   "maxMemory": 7456.948224,
   "freeMemory": 294.53928,
   "totalMemory": 524.288,
   "memoryInfo": "229.7487MB / 524.2880MB (7456.9482MB Max)"
}
```

The raw memory values are not limited to their decimal places, the memoryInfo (formatted data) is limited to 4 decimal places each (Defined in RuntimeInfo L24)

Kichura (approved):

PierreSchwang:
@Kichura I have mergedthe changes since the creation of the PR because you approved

kennytv:
Memory information isn't actually too useful

PR #2494, opened 2021-05-25
Build script improvements

jpenilla:
Moves most shared logic into convention plugins instead of cross-configuring everything from the main build script.


KennyTV (approved):
Thank you jomp, very cool

PR #2528, opened 2021-06-10
Improve arm-swing cancel by resetting cursor item

tomcraft:
Improvements of #2031 and #2007 , based on @Gerrygames [comment](https://github.com/ViaVersion/ViaVersion/pull/2031#issuecomment-678677104) and help with his [code snippet on Discord](https://discord.com/channels/316206679014244363/316218802155028482/852582496986333254).

As @Gerrygames said in his comment and on Discord, the source of the issue is that the client tries to drop the item in cursor during navigation between multiple inventories. This was causing an arm swing animation packet to be sent and processed by server as an interaction.

Previous PR was cancelling any arm animation packet while an inventory is open. This required tracking inventories open and close and the arm was this swinging client-side. It worked well with 1.16 but with 1.17 the packet order changed and the animation packet is now sent after the inventory closed, causing the issue to reappear. So it is now impossible to track for inventory state for 1.17.

This PR clears the cursor item before `WINDOW_OPEN` and before `WINDOW_CLOSE` to prevent client to try dropping any item. Clearing at `WINDOW_OPEN` prevents drop during navigation, and `WINDOW_CLOSE` prevents drop at the very end when a click on an item closes the inventory. As a logical effect, there is no more client-side arm swing.

I also removed the now unused InventoryTracker that was introduced to track windowId.

All feedbacks are welcome :)


tomcraft:
just fixed the way i registered remappers

RednedEpic:
LGTM, definitely seems to be a more proper fix for the issue.

connorhartley:
I agree, LGTM.

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_16to1_15_2/packets/InventoryPackets.java
```diff
@@ -45,6 +47,16 @@ public InventoryPackets(Protocol1_16To1_15_2 protocol) {
 
     @Override
     public void registerPackets() {
+        // clear cursor item to prevent client to try dropping it during navigation between multiple inventories causing arm swing
+        PacketHandler cursorRemapper = wrapper -> {
+            PacketWrapper clearPacket = wrapper.create(ClientboundPackets1_16.SET_SLOT);
+            clearPacket.write(Type.BYTE, (byte)-1);
```
KennyTV, line 53: the type was changed 1/2 days ago
```suggestion
            clearPacket.write(Type.UNSIGNED_BYTE, (byte)-1);
```

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_16to1_15_2/packets/InventoryPackets.java
```diff
@@ -45,6 +47,16 @@ public InventoryPackets(Protocol1_16To1_15_2 protocol) {
...
+            clearPacket.write(Type.SHORT, (short)-1);
+            clearPacket.write(Type.FLAT_VAR_INT_ITEM, null);
+            clearPacket.send(Protocol1_16To1_15_2.class);
+            wrapper.passthroughAll();
```
KennyTV, line 57: that line isn't needed

KennyTV (changes requested):
Looks good otherwise

tomcraft:
Applied kneny's suggestions

KennyTV (approved):

PR #2561, opened 2021-06-20
use a more fork-proof approach to enable Paper block place patch

tomcraft:
tested with:
 - Paper 1.8.8, 1.9.4, 1.11.2 (enabling)
 - SportPaper 1.8.8 (enabling)
 - Spigot 1.8.8 (not enabling)

### DIFF
--- a/bukkit-legacy/src/main/java/com/viaversion/viaversion/bukkit/listeners/protocol1_9to1_8/PaperPatch.java
+++ b/bukkit-legacy/src/main/java/com/viaversion/viaversion/bukkit/listeners/protocol1_9to1_8/PaperPatch.java
@@ -19,6 +19,7 @@ package com.viaversion.viaversion.bukkit.listeners.protocol1_9to1_8;
 
 import org.bukkit.Location;
 import org.bukkit.Material;
+import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.event.EventHandler;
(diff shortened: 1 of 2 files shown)


Xernium (approved):

FormallyMyles:
Code looks great, however, do we also need a way to disable the patch? Given that some forks can remove the faulty code.

tomcraft:
I added a system property to disable this patch, is it an ok-enough approach ?

FormallyMyles:
> I added a system property to disable this patch, is it an ok-enough approach ?

I don't see any issue with that, also means the fork could control it if they wanted to

FormallyMyles (approved):

andreasdc:
Oh no, what if someone doesn't want this patch? It's buggy.

kennytv:
Then you put the system property `com.viaversion.ignorePaperBlockPlacePatch` as `true`

andreasdc:
Oh yes I didn't see this commit, It would be cool to have that in config too.

andreasdc:
So basically '-com.viaversion.ignorePaperBlockPlacePatch=true' in flags?

tomcraft:
-Dcom.viaversion.ignorePaperBlockPlacePatch=true

andreasdc:
I think having it in the config would be better, also I don't think so you want to add this patch, without additional changes in it creates many bugs.

PR #2600, opened 2021-07-03
reduce World#isChunkLoaded usage for cached Chunk instance

tomcraft:
i think it is safe to ignore this check for the `lastChunk` Chunk instance, i've been running a similar behavior implementation based on NMS for many month without issue.

feel free to comment :)


kennytv (changes requested):
Nope, the chunk might have been unloaded already.

tomcraft:
Oke, I don't see much more to do here without big changes.

andreasdc:
@kennytv what's the point in caching lastChunk then?

kennytv:
Because this reduces the risk of issues (like cancelling unloads) due to concurrency issues, which block connection methods are still full of.

andreasdc:
You are right, but 1 single getChunkIfLoaded would be better right?

kennytv:
That doesn't exist in these old versions, and we are not interested in doing bytecode hacks or using version dependent reflection to get them.

andreasdc:
That doesn't answer my question, but I understand, thanks. :)

andreasdc:
@kennytv isn't this method called on loaded chunks? Like placing blocks etc.?

PR #2627, opened 2021-07-16
experimental idle packet rework

creeper123123321:


creeper123123321:
seems to help to reduce anticheat triggering

creeper123123321:
todo check for the 1s idle packet

PR #2631, opened 2021-07-19
fix wrong ticking in proxies

creeper123123321:


kennytv (changes requested):
Reuse `userConnection.get(MovementTracker.class)` from above

PR #2672, opened 2021-08-29
Add support for setting a custom world name

Levertion:
This applies for 1.15 or below servers with 1.16 or above clients.

I'm contributing with the hopes that [Shotbow](https://shotbow.net/) will use it - their developers have expressed interest in their discord.

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/minecraft/WorldIdentifiers.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/minecraft/WorldIdentifiers.java
@@ -3,21 +3,33 @@ package com.viaversion.viaversion.api.minecraft;
 import com.viaversion.viaversion.api.connection.StorableObject;
 
 public class WorldIdentifiers implements StorableObject {
-    public static String overworldDefault = "minecraft:overworld";
-    public static String netherDefault = "minecraft:the_nether";
-    public static String endDefault = "minecraft:the_end";
+    public static final String OVERWORLD_DEFAULT = "minecraft:overworld";
+    public static final String NETHER_DEFAULT = "minecraft:the_nether";
+    public static final String END_DEFAULT = "minecraft:the_end";
 
-    public String overworld;
-    public String nether;
-    public String end;
+    private final String overworld;
+    private final String nether;
+    private final String end;
 
-    public WorldIdentifiers(String overworld){
-        this(overworld, netherDefault, endDefault);
+    public WorldIdentifiers(String overworld) {
+        this(overworld, NETHER_DEFAULT, END_DEFAULT);
     }
 
-    public WorldIdentifiers(String overworld, String nether, String end){
+    public WorldIdentifiers(String overworld, String nether, String end) {
         this.overworld = overworld;
         this.nether = nether;
         this.end = end;
+    }
+
+    public String overworld() {
+        return this.overworld;
+    }
+
+    public String nether() {
+        return this.nether;
+    }
+
+    public String end() {
+        return this.end;
     }
 }
--- a/api/src/main/java/com/viaversion/viaversion/api/minecraft/WorldIdentifiers.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/minecraft/WorldIdentifiers.java
@@ -1,0 +1,23 @@
+package com.viaversion.viaversion.api.minecraft;
+
+import com.viaversion.viaversion.api.connection.StorableObject;
+
+public class WorldIdentifiers implements StorableObject {
+    public static String overworldDefault = "minecraft:overworld";
+    public static String netherDefault = "minecraft:the_nether";
+    public static String endDefault = "minecraft:the_end";
+
+    public String overworld;
+    public String nether;
+    public String end;
+
+    public WorldIdentifiers(String overworld){
+        this(overworld, netherDefault, endDefault);
+    }
+
+    public WorldIdentifiers(String overworld, String nether, String end){
+        this.overworld = overworld;
+        this.nether = nether;
+        this.end = end;
+    }
+}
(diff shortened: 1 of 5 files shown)


kennytv:
What exactly is the benefit of doing that?

Levertion:
So specifically it would be useful for mods which want to know which world you're in for a multiverse server with a lower than 1.16 base version.

For example, a minimap mod should display a seperate set of waypoints depending on the current world. At the moment, on shotbow my waypoints for different worlds overlap.
It would be possible to do hacky detection and conversion in a client side mod, but it would be much better if the server were able to report it properly itself.

common/src/main/java/com/viaversion/viaversion/configuration/AbstractViaConfig.java
```diff
@@ -141,6 +145,15 @@ protected void loadFields() {
         ignoreLongChannelNames = getBoolean("ignore-long-1_16-channel-names", true);
         forcedUse1_17ResourcePack = getBoolean("forced-use-1_17-resource-pack", false);
         resourcePack1_17PromptMessage = getSerializedComponent("resource-pack-1_17-prompt");
+        map1_16WorldNames.clear();
+        JsonElement maybeMap = getSerializedComponent("map-1_16-world-names");
```
kennytv, line 149: This is not a component... Unsure if the Via config even supports subsections

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_16to1_15_2/packets/EntityPackets.java
```diff
@@ -201,7 +200,7 @@ public void registerMap() {
                 map(Type.UNSIGNED_BYTE); //  Gamemode
                 handler(wrapper -> {
                     wrapper.write(Type.BYTE, (byte) -1); // Previous gamemode, set to none
-                    wrapper.write(Type.STRING_ARRAY, WORLD_NAMES); // World list - only used for command completion
+                    wrapper.write(Type.STRING_ARRAY, Via.getConfig().get1_16WorldNamesMap().values().toArray(new String[0])); // World list - only used for command completion
```
kennytv, line 203: Remove this, as it is basically useless with servers providing their own completion prior to the change

kennytv (changes requested):
Makes sense, thanks; still some changes to do.

common/src/main/java/com/viaversion/viaversion/configuration/AbstractViaConfig.java
Levertion: Is the actual shape of the config fine?

You are correct, the config item should already be a `Map<String, String>` which should be easily castable.

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_16to1_15_2/packets/EntityPackets.java
Levertion: Right, so clients don't really care about it being correct, since it's only use is for the specific brigadier `world` parser?

Removed.

api/src/main/java/com/viaversion/viaversion/api/minecraft/WorldIdentifiers.java
```diff
@@ -0,0 +1,23 @@
+package com.viaversion.viaversion.api.minecraft;
+
+import com.viaversion.viaversion.api.connection.StorableObject;
+
+public class WorldIdentifiers implements StorableObject {
+    public static String overworldDefault = "minecraft:overworld";
+    public static String netherDefault = "minecraft:the_nether";
+    public static String endDefault = "minecraft:the_end";
```
kennytv, line 8: `public static final` and upper case names (`OVERWORLD_DEFAULT` etc.)

api/src/main/java/com/viaversion/viaversion/api/minecraft/WorldIdentifiers.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public String overworld;
+    public String nether;
+    public String end;
```
kennytv, line 12: `private final`, add getters

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_16to1_15_2/packets/EntityPackets.java
```diff
@@ -42,25 +43,35 @@
 public class EntityPackets {
 
     private static final PacketHandler DIMENSION_HANDLER = wrapper -> {
+        WorldIdentifiers map = Via.getConfig().get1_16WorldNamesMap();
+        WorldIdentifiers userMap = wrapper.user().get(WorldIdentifiers.class);
+        if (userMap!=null){
+            map = userMap;
+        }
```
kennytv, line 50: Is there actually a need for per player world identifiers for other plugins to set?

kennytv (commented):
Just some minor things to do, ready to good after that

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_16to1_15_2/packets/EntityPackets.java
Levertion: I think so - that was the primary motivation for this second set of changes. For example, in Shotbow, one of their game modes has the 'lobby' and the actual map - and you get sent back to the lobby when you die.
As far as I'm aware, these are on the same server with the same ViaVersion version, but not all players are in the same one.

Kichura (commented):
This commit made viaversion fork not compilable, perhaps resolve that?

Levertion:
That is a very reasonable suggestion - done, I think.

Kichura (approved):
Seems to compile properly now.

kennytv (approved):
Thank you

PR #2684, opened 2021-09-12
Load protocols only on demand

lixfel:
**Rationale**: ViaVersion takes a long time (on my server ~1.5s) to initialize and likely initializes protocols never required.

This branch was tested on a Spigot 1.15.2 server behind a Waterfall proxy with Client versions 1.12.2, 1.15.2 and 1.17.1. Tests were done with and without ProtocolLib and ViaBackwards.

Calling the registerProtocol method still initializes the protocol (and all required protocols) due to the ProtocolManager.registerProtocol documentation. This could be changed (it was possible to join with 1.12.2 and ViaBackwards without enforcing initialization during registration). The repeating task for checking running mappingloaders and potential json cache clear is debatable.

I can think of the following (untested) potential issues with this implementation:
- (Main)Thread blocking during delayed initialization on other platforms than Spigot
- Bad ping during first server list ping with uninitialized protocol version

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/ViaManagerImpl.java
+++ b/common/src/main/java/com/viaversion/viaversion/ViaManagerImpl.java
@@ -57,7 +57,6 @@ public class ViaManagerImpl implements ViaManager {
     private final ViaPlatformLoader loader;
     private final Set<String> subPlatforms = new HashSet<>();
     private List<Runnable> enableListeners = new ArrayList<>();
-    private PlatformTask mappingLoadingTask;
     private boolean debug;
 
     public ViaManagerImpl(ViaPlatform<?> platform, ViaInjector injector, ViaCommandHandler commandHandler, ViaPlatformLoader loader) {
@@ -153,12 +152,7 @@ public class ViaManagerImpl implements ViaManager {
         // Load Platform
         loader.load();
         // Common tasks
-        mappingLoadingTask = Via.getPlatform().runRepeatingSync(() -> {
-            if (protocolManager.checkForMappingCompletion()) {
-                mappingLoadingTask.cancel();
-                mappingLoadingTask = null;
-            }
-        }, 10L);
+        Via.getPlatform().runRepeatingSync(protocolManager::checkForMappingCompletion, 10L);
 
         int serverProtocolVersion = protocolManager.getServerProtocolVersion().lowestSupportedVersion();
         if (serverProtocolVersion < ProtocolVersion.v1_9.getVersion()) {
--- a/common/src/main/java/com/viaversion/viaversion/ViaManagerImpl.java
+++ b/common/src/main/java/com/viaversion/viaversion/ViaManagerImpl.java
@@ -20,7 +20,6 @@ package com.viaversion.viaversion;
 import com.viaversion.viaversion.api.Via;
 import com.viaversion.viaversion.api.ViaManager;
 import com.viaversion.viaversion.api.connection.ConnectionManager;
-import com.viaversion.viaversion.api.platform.PlatformTask;
 import com.viaversion.viaversion.api.platform.UnsupportedSoftware;
 import com.viaversion.viaversion.api.platform.ViaInjector;
 import com.viaversion.viaversion.api.platform.ViaPlatform;
@@ -151,9 +150,8 @@ public class ViaManagerImpl implements ViaManager {
 
         // Load Platform
         loader.load();
-        // Common tasks
-        Via.getPlatform().runRepeatingSync(protocolManager::checkForMappingCompletion, 10L);
 
+        // Common tasks
         int serverProtocolVersion = protocolManager.getServerProtocolVersion().lowestSupportedVersion();
         if (serverProtocolVersion < ProtocolVersion.v1_9.getVersion()) {
             if (Via.getConfig().isSimulatePlayerTick()) {
--- a/common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
@@ -98,8 +98,8 @@ public class ProtocolManagerImpl implements ProtocolManager {
     private static final Protocol BASE_PROTOCOL = new BaseProtocol();
 
     // Input Version -> Output Version & Protocol (Allows fast lookup)
-    private final Int2ObjectMap<Int2ObjectMap<ProtocolSupplier>> registryMap = new Int2ObjectOpenHashMap<>(32);
-    private final Map<Class<? extends Protocol>, ProtocolSupplier> protocols = new HashMap<>();
+    private final Int2ObjectMap<Int2ObjectMap<Supplier<Protocol>>> registryMap = new Int2ObjectOpenHashMap<>(32);
+    private final Map<Class<? extends Protocol>, Supplier<Protocol>> protocols = new HashMap<>();
     private final Map<ProtocolPathKey, List<ProtocolPathEntry>> pathCache = new ConcurrentHashMap<>();
     private final Set<Integer> supportedVersions = new HashSet<>();
     private final List<Pair<Range<Integer>, Protocol>> baseProtocols = Lists.newCopyOnWriteArrayList();
@@ -171,7 +171,8 @@ public class ProtocolManagerImpl implements ProtocolManager {
 
     @Override
     public void registerProtocol(Protocol protocol, List<Integer> supportedClientVersion, int serverVersion) {
-        registerProtocolSupplier(protocol.getClass(), new DirectProtocolSupplier(protocol), supportedClientVersion, serverVersion);
+        setupProtocolSupplier(protocol.getClass(), () -> protocol, supportedClientVersion, serverVersion);
+        initializeProtocol(protocol);
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
```diff
@@ -487,28 +479,45 @@ public void onServerLoaded() {
...
-
     private Function<Throwable, Void> mappingLoaderThrowable(Class<? extends Protocol> protocolClass) {
         return throwable -> {
             Via.getPlatform().getLogger().severe("Error during mapping loading of " + protocolClass.getSimpleName());
             throwable.printStackTrace();
             return null;
         };
     }
+
+    private class CachedProtocolSupplier implements Supplier<Protocol> {
```
kennytv, line 490: final

common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
```diff
@@ -487,28 +479,45 @@ public void onServerLoaded() {
...
+        private final Supplier<Protocol> provider;
+        private Protocol protocol = null;
```
kennytv, line 492:
```suggestion
        private Protocol protocol;
```

common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
```diff
@@ -487,28 +479,45 @@ public void onServerLoaded() {
...
+
+        public CachedProtocolSupplier(Supplier<Protocol> provider) {
+            this.provider = provider;
+        }
+
+        @Override
+        public Protocol get() {
+            synchronized (provider) {
+                if (protocol == null) {
```
kennytv, line 501: Early return if protocol != null and continue without indenting otherwise

common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
```diff
@@ -458,9 +440,19 @@ public void addMappingLoaderFuture(Class<? extends Protocol> protocolClass, Clas
 
     @Override
     public @Nullable CompletableFuture<Void> getMappingLoaderFuture(Class<? extends Protocol> protocolClass) {
+        CompletableFuture<Void> future = internalGetMappingLoaderFuture(protocolClass);
+        if (future != null)
```
kennytv, line 444: Add brackets

common/src/main/java/com/viaversion/viaversion/ViaManagerImpl.java
```diff
@@ -153,12 +152,7 @@ public void onServerLoaded() {
         // Load Platform
         loader.load();
         // Common tasks
-        mappingLoadingTask = Via.getPlatform().runRepeatingSync(() -> {
-            if (protocolManager.checkForMappingCompletion()) {
-                mappingLoadingTask.cancel();
-                mappingLoadingTask = null;
-            }
-        }, 10L);
+        Via.getPlatform().runRepeatingSync(protocolManager::checkForMappingCompletion, 10L);
```
kennytv, line 155: This doesn't look right. The mappings cache use is a bit strange, admittedly, but I'll have to double check

common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
```diff
@@ -176,39 +171,32 @@ public void registerProtocol(Protocol protocol, ProtocolVersion clientVersion, P
 
     @Override
     public void registerProtocol(Protocol protocol, List<Integer> supportedClientVersion, int serverVersion) {
-        // Register the protocol's handlers
-        protocol.initialize();
+        registerProtocolSupplier(protocol.getClass(), () -> protocol, supportedClientVersion, serverVersion);
+        getProtocol(protocol.getClass());  // Api guarantees initialisation
```
kennytv, line 175: No need to call getProtocol here if there's no actual initialisation anyways - but to make its use clearer and avoid unnecessary synchronization, I'd suggest the following: Make CachedProtocolSupplier an interface and add a second implementation only holding a final protocol that is directly returned, which is then used here

kennytv (changes requested):
Some minor suggestions without having tested the code. If you haven't already, please make sure this works with ViaVersion alone, and combined with ViaBackwards (and ViaRewind) on multiple server versions - 1.8, 1.12, 1.17 should be enough with any client version other than the server version

common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
lixfel: Did change it in an abstract class instead of an interface due to member specific function calls not included in the ProtocolManager-API in the initialisation code (I consider loading the mappings, protocol.initialize and protocol.register as some sort of initialization).

lixfel:
Tested now also with 1.8.9 Paper - 1.15.2 Client, 1.12.2 Spigot - 1.15.2 Client and with 1.17 Spigot - 1.15.2 Client (with ViaBackwards + ViaRewind). A test run with 1.17 Spigot - 1.8.9 Client did fail due to a missing "Chunk1_8" class (which I could not find in the current repository), the same test run with unmodified ViaVersion 4.1 did work.

tomcraft:
> Tested now also with 1.8.9 Paper - 1.15.2 Client, 1.12.2 Spigot - 1.15.2 Client and with 1.17 Spigot - 1.15.2 Client (with ViaBackwards + ViaRewind). A test run with 1.17 Spigot - 1.8.9 Client did fail due to a missing "Chunk1_8" class (which I could not find in the current repository), the same test run with unmodified ViaVersion 4.0.1 did work.

you need latest Rewind/Backwards builds, not the ones in spigot website.

kennytv:
After thinking about this, I think this can be simplified by a bit if you lazy load the mapping data, not the protocols. Protocol registration and creation is basically free; the thing doing all the data caching and putting temporary strain on cpu is the mapping loading and i/o (done in the `addMappingLoaderFuture` methods). You could probably implement a similar change to the `mappingLoaderFutures` map to either hold the started CompletableFuture or the Runnable to execute later.

Also, just remove the `checkForMappingCompletion` runnable for now, I will deal with that to a greater extent later. Keeping the string data cached doesn't hurt all that much in the meantime.

kennytv:
When you have a great number of (virtual) threads suddenly taking a toll on your cpu because of about a dozen different files being loaded and processed while the protocols themselves are being loaded, that is obviously going to affect that as well. I can guarantee you that if you simply disable mapping loading and fill the future map with `.completedFuture(null)`, the protocol inits will go down to less than a 10th of a second (if you exclude the profiler itself taking its time, especially if not using an async profiler), take or give a few - which does not warrant this level of extra complexity.

lixfel:
Completly disabling mapping loading (mappingsLoaded = true, early return in both addMappingLoaderFuture) on current master did reduce loading time, but only to 762ms (ViaVersion.onLoad+ViaBackwards.onLoad). I'm using VisualVM in sampling mode, so performance impact is very small (have not been able to measure the impact).

kennytv:
Changing that boolean does not disable mapping loading

lixfel:
But immediately returning in both addMappingLoaderFeature functions does (no MappingLoader-Threads have been spawned).

kennytv:
My previous comment was wrong, I misread your message, but in any case: Premature optimization is bad, especially if you don't know what exactly is slow. 500 of those 700ms might as well come from the files ViaBackwards always loads anyways, aside from mapping loading. In the way it currently is, I would like to see proper async profiling done to confirm it is actually protocol initialization that takes that long (taking into account VB's force loading of two files), and also what *exactly* is slow.

lixfel:
I agree that premature optimization is bad, but after especially looking bottom up at some sampling (aka async profiling) results I don't think that the content of this branch is premature optimization.
grafik
(Sampling result vanilla ViaVersion 4.0.1 and ViaBackwards 4.0.1; only main server thread; only Via* loading time (system.in.read() prior and post plugin loading); sampling frequency 1ms)
grafik
(Sampling result with deactivated mapping loading ViaVersion master and ViaBackwards 4.0.1; only main server thread; only Via* loading time (system.in.read() prior and post plugin loading); sampling frequency 1ms)

What I read from my sampling results:
- There is no big bottleneck, the initialisation is pretty distributed
- A bottleneck is the Spigot PluginClassLoader nonsense (they seem to check all classes for usage of pre flattening materials etc.); I think this is out of scope here
- ViaBackwards always loaded files do not matter much (only 17ms between cpu and total time with mapping loading, else always 0ms), ViaBackwards and ViaVersion always have a similar load time

kennytv:
Thanks for doing the profiling. After some internal discussion, I can understand why the slightly higher startup time may annoy you, but simply postponing 800ms of initialization to the first poor soul that tries to connect with latest/oldest client version is definitely not the way to go, given it's largely out of our hands to reduce the time with the platform's class loading (other, smaller parts we may improve in the future).

In the grand scheme of things, 400/800ms startup time is more or less nothing, relative to what Via actually does in the end. If you have three dozen plugins taking a second each, then it's probably up for those to save on time, but simply is not worth the maintenance and negative side effects in Via. If this concerns wanting faster startup time for frequently restarted test servers, you might consider removing Via or parts of Via you don't need during testing.

We'd still see some benefit to lazily loading mappings, since it splits up workload and also saves a few of those precious megabytes of ram and i/o, but lazy loading protocols is not something we see much benefit in.

lixfel:
Okay, then I will have to maintain my own patches, what I would have liked to avoid. For me, the 1.5s startup time is a significant portion of my already optimized startup time (~7s) and if the player have to wait prior join or (likly shorter) during join is a difference for me.

PR #2698, opened 2021-10-02
Remove usage of `sun.misc.Unsafe` in `BungeeViaInjector`

### LINKED ISSUE #2414: Bungee Java 16 compatibility and injector cleanup
The Bungee injector has stopped working with Java 16 given final field modification hacks. We need to either somehow inject into th child handler in the `ServerBootstrap` instances or pr a small hack to Bungee/Waterfall for injection without reflection.

In general, all injectors could use a little cleanup. ☠

Check-list:

- [x] 1.16 Compatibility
- [ ] Injector clean-up

MrMicky-FR:
This removes the usage of `sun.misc.Unsafe` in `BungeeViaInjector`, by injecting in the child handler, like in the Bukkit/Sponge injectors, which doesn’t require to modify a static final field, so it doesn’t require the usage of `sun.misc.Unsafe`, which may break in future Java versions.

I tested with only ViaVersion, with ViaVersion/ViaBackwards and with ViaVersion/ViaBackwards/ViaRewind, on both BungeeCord and Waterfall, with newer and older clients, and everything seems to works fine, so I guess it’s good 😃 

This should also closes #2414.

### DIFF
--- a/bungee/src/main/java/com/viaversion/viaversion/bungee/platform/BungeeViaInjector.java
+++ b/bungee/src/main/java/com/viaversion/viaversion/bungee/platform/BungeeViaInjector.java
@@ -17,46 +17,54 @@
  */
 package com.viaversion.viaversion.bungee.platform;
 
+import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import com.viaversion.viaversion.api.Via;
 import com.viaversion.viaversion.api.platform.ViaInjector;
 import com.viaversion.viaversion.bungee.handlers.BungeeChannelInitializer;
-import com.viaversion.viaversion.compatibility.ForcefulFieldModifier;
-import com.viaversion.viaversion.compatibility.unsafe.UnsafeBackedForcefulFieldModifier;
 import com.viaversion.viaversion.util.ReflectionUtil;
+import com.viaversion.viaversion.util.SetWrapper;
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelInitializer;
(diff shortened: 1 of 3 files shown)


bungee/src/main/java/com/viaversion/viaversion/bungee/platform/BungeeViaInjector.java
```diff
@@ -17,46 +17,54 @@
...
+        });
+
+        this.listenersField.set(server, wrapper);
 
-            this.forcefulFieldModifier.setField(field, null, newInit);
-        } catch (Exception e) {
-            Via.getPlatform().getLogger().severe("Unable to inject ViaVersion, please post these details on our GitHub and ensure you're using a compatible server version.");
-            throw e;
+        // Iterate through current list
+        synchronized (wrapper) {
```
kennytv, line 64: No need to synchronize here if done nowhere else

bungee/src/main/java/com/viaversion/viaversion/bungee/platform/BungeeViaInjector.java
```diff
@@ -17,46 +17,54 @@
...
-
-    public BungeeViaInjector() {
-        try {
-            this.forcefulFieldModifier = new UnsafeBackedForcefulFieldModifier();
-        } catch (final ReflectiveOperationException ex) {
-            throw new IllegalStateException("Cannot create a modifier accessor", ex);
-        }
-    }
+    private final List<Channel> injectedChannels = new ArrayList<>();
+    private Field listenersField;
```
kennytv, line 42: Assign this in a static initializer

Issue #2187, opened 2020-11-12
BUG: Protocol ossification in plugin channel names


creeper123123321:
https://en.wikipedia.org/wiki/Protocol_ossification
https://github.com/ViaVersion/ViaVersion/blob/692e53427b41f3f564e4e5701c38fa9b7553e7a0/common/src/main/java/us/myles/ViaVersion/protocols/protocol1_13to1_12_2/data/MappingData.java#L130
Fails to match these valid identifiers: "no_namespace_but_valid", "empty_path_but_valid:"
Empty namespace (like ":example") is parsed as "minecraft" ("minecraft:example")

creeper123123321:
See #1448

creeper123123321:
Prismarine web client uses "brand" with no "minecraft:"

Gerrygames:
Is ``:`` a valid identifier?

PR #2701, opened 2021-10-03
Fix channel name validation in 1.13 to 1.12.2 protocol

Gerrygames:
As partially described in #2187 the following channel names are valid but not recognized as valid:
- an empty string
- ``:``
- ``no_namespace`` (parsed as minecraft namespace)
- ``empty_path:``
- ``:empty_namespace`` (parsed as minecraft namespace)

This PR fixes all of these cases.

This PR also fixes a different issue which couldn't occur at runtime due to the first issue.
The ``protocol1_13to1_12_2.data.MappingData#validateNewChannel`` method would transform ``:path`` to ``minecraft::path`` which is invalid.
I did also remove the length check in that method because it was not necessary. The only place where this method is used does limit the length itself:
https://github.com/ViaVersion/ViaVersion/blob/e64a0fb62e8573096bee85581c2e5f263654591d/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/InventoryPackets.java#L679-L703
The removed length check would only result in ``debug/neighbors_update`` falling through the switch case here.

I did check if everything is still working with a 1.12.2 server and 1.17.1 client. Everything is working fine.
I added the examples listed above to ``channelmappings-1.13.json`` and all of them are recognized as valid.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/data/MappingData.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/data/MappingData.java
@@ -147,7 +147,7 @@ public class MappingData extends MappingDataBase {
     }
 
     public static boolean isValid1_13Channel(String channelId) {
-        return channelId.matches("[0-9a-z_.-]*:?[0-9a-z_/.-]*");
+        return channelId.matches("([0-9a-z_.-]+:)?[0-9a-z_/.-]+");
     }
 
     private void loadTags(Map<String, Integer[]> output, JsonObject newTags) {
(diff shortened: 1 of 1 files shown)


kennytv (changes requested):
This needs extra checks on Spigot servers specifically, maybe Bungee?, else you get kicked there due to less lenient restrictions

Gerrygames:
I don't think this needs any special handling for Spigot or any other server software. The new channel names are only between ViaVersion and the client. The server is 1.12.2 or older in this case and doesn't have any restrictions besides the length limit.
I will check if bungee has any restrictions tho.

kennytv:
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/browse/src/main/java/org/bukkit/NamespacedKey.java?at=f802d0a5711b4497c7d1c92108a18b34482f5fc2#32

Gerrygames:
Okay I will change to only accept ``namespace:path`` and ``path`` for now. Those are the most sensible formats and it will fix clients sending ``brand`` instead of ``minecraft:brand``.

kennytv (approved):

PR #2745, opened 2021-12-02
Map smoke directions from 1.13 to 1.13.1

WinX64:
There has been a change in directions of the smoke effect back in 1.13.1 that has gone relatively unnoticed. Below is a comparison from before and after (including the data of each direction):
image

Some of the old directions are no longer reproducible as of 1.13.1, but North, South, East and West are, and this PR maps them to their new data values. For the old unsupported ones, I added a fallback to 0 (Down), or should it be something else?

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13_1to1_13/Protocol1_13_1To1_13.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13_1to1_13/Protocol1_13_1To1_13.java
@@ -144,6 +144,41 @@ public class Protocol1_13_1To1_13 extends AbstractProtocol<ClientboundPackets1_1
             }
         });
 
+        registerClientbound(ClientboundPackets1_13.EFFECT, new PacketRemapper() {
+            @Override
+            public void registerMap() {
+                map(Type.INT); // Effect Id
+                map(Type.POSITION); // Location
+                map(Type.INT); // Data
+                handler(new PacketHandler() {
+                    @Override
+                    public void handle(PacketWrapper wrapper) throws Exception {
+                        int id = wrapper.get(Type.INT, 0);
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13_1to1_13/Protocol1_13_1To1_13.java
```diff
@@ -144,6 +144,41 @@ public void handle(PacketWrapper wrapper) throws Exception {
...
+            @Override
+            public void registerMap() {
+                map(Type.INT); // Effect Id
+                map(Type.POSITION); // Location
+                map(Type.INT); // Data
+                handler(new PacketHandler() {
+                    @Override
+                    public void handle(PacketWrapper wrapper) throws Exception {
+                        int id = wrapper.get(Type.INT, 0);
+                        int data = wrapper.get(Type.INT, 1);
```
Camotoy, line 157: Small optimization - move this get call to below the if statement on the next line since it's not needed until then?

kennytv:
Having 0 2 6 8 map to something close (i.e. 0 2 => 5 and 6 8 => 4) would be nice. Also, could you provide a few simple steps to test this out? Not sure how that effect can be created

WinX64:
> Having 0 2 6 8 map to something close (i.e. 2 8 => 5 and 6 0 => 4) would be nice. Also, could you provide a few simple steps to test this out? Not sure how that effect can be created

Sure. On Bukkit, a simple call to `Player#playEffect(<location>, Effect.SMOKE, <data>)` should do the trick. Test it with all the data values (0 to 8), a small delay between them, on both 1.13 and 1.13.1, and you should get the same results I described.

On 1.13.1, it should look something like this (head is facing North):
SmokeDirections

kennytv (approved):
Thank you!

PR #2795, opened 2022-01-09
Allow "No light data found" message to be suppressed

### LINKED ISSUE #2791: Add config item to disable "No light data found..." message
### Problem Description

I got an explanation from the Orebfuscator developers of how they think the light data problem is occurring:

> Problem is that we don't touch the light packets, we only work with chunk packets. But chunk packets are handled async which means they are sent independently of the light packets... What probably happens is the following:
> 1. server sends light
> 2. sever sends chunk
> 3. _orebfuscator starts processing chunk packet_
> 4. server sends unload chunk
> 5. _orebfuscator sends processed chunk packet_
>
> This would cause ViaVersion to store the light data and delete it on chunk unload. Which in turn causes the delayed chunk packet to not contain any light data. Since the chunk is already unloaded this shouldn't cause any problems for your players but the error will remain.

### Solution Description

Add an item in the config file to disable the "No light data found for chunk at... Chunk was already loaded..." warning message.

### Alternatives

The Orebfuscator developers say they would have to delay the chunk unload if a chunk is still processing, which to me sounds like it could cause more problems than it would solve.

Issue on Orebfuscator's Github: https://github.com/Imprex-Development/Orebfuscator/issues/165#issuecomment-1003389772

### Additional Info

Problem occurs with 1.18.x clients on a 1.17.1 server.

Bobcat00:
This uses the existing suppress-conversion-warnings config entry to suppress the
"No light data found for chunk at... Chunk was already loaded..." warning.
This message can occur if a plugin does asynchronous chunk processing and a
chunk is unloaded before the plugin's processing is completed. The warning
occurs frequently when a 1.17.1 server is used with a 1.18.1 client.

Fixes #2791

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_18to1_17_1/packets/WorldPackets.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_18to1_17_1/packets/WorldPackets.java
@@ -168,7 +168,9 @@ public final class WorldPackets {
                     final ChunkLightStorage.ChunkLight light = Via.getConfig().cache1_17Light() ?
                             lightStorage.getLight(chunk.getX(), chunk.getZ()) : lightStorage.removeLight(chunk.getX(), chunk.getZ());
                     if (light == null) {
-                        Via.getPlatform().getLogger().warning("No light data found for chunk at " + chunk.getX() + ", " + chunk.getZ() + ". Chunk was already loaded: " + alreadyLoaded);
+                        if (!Via.getConfig().isSuppressConversionWarnings() || Via.getManager().isDebug()) {
+                            Via.getPlatform().getLogger().warning("No light data found for chunk at " + chunk.getX() + ", " + chunk.getZ() + ". Chunk was already loaded: " + alreadyLoaded);
+                        }
 
                         final BitSet emptyLightMask = new BitSet();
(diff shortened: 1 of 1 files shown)


creeper123123321 (approved):

kennytv:
If Via receives it in this order, that means the client will receive it in that order - which is bad and shouldn't happen, this warning isn't one that should be suppressed

Bobcat00:
As I understand it, Via normally gets the data in order. However, if Via gets the light data but then the chunk gets unloaded before Via gets the chunk data, Via complains when it gets the chunk data for the unloaded chunk.

kennytv:
... which means the client will load the chunk again because it receives the chunk data after the unload (which it shouldn't)

Bobcat00:
Please see this comment: https://github.com/Imprex-Development/Orebfuscator/issues/165#issuecomment-1003389772

kennytv:
The same argument I made still stands; not a Via issue, and definitely not something you can just ignore

Bobcat00:
I just want to get rid of the _repeated spamming_ of the message.

kennytv:
You can change your logging configuration to filter it or get another plugin doing that. Unlike other cases the suppress config option is for, this is a valid and important warning we won't just hide

Bobcat00:
I'm getting 10 messages per minute and they're meaningless.

Issue #2835, opened 2022-03-03
Race condition on PlayerConnection initialization, causing ViaAPI to report wrong information for connecting players


Pablete1234:
### `/viaversion dump` Output

https://dump.viaversion.com/c9cc443fbfb4e3133b8f747a589fee15f9bbcd6056df666fc811ee5d4c1c59b0

### Bug Description

`ViaAPI#getPlayerVersion` can return -1 on PlayerJoinEvent, depending on server capacity (it is caused by a race condition)

Relevant codepaths:

`ViaAPI#getPlayerVersion` relies down to `ConnectionManager#getConnectedClient(uuid)`: https://github.com/ViaVersion/ViaVersion/blob/master/bukkit/src/main/java/com/viaversion/viaversion/bukkit/platform/BukkitViaAPI.java#L45
The `ConnectionManager`, will only have a player to return from `getConnectedClient` if `onLoginSuccess` has been called for that user, since that is what adds it to the map:
https://github.com/ViaVersion/ViaVersion/blob/master/common/src/main/java/com/viaversion/viaversion/connection/ConnectionManagerImpl.java#L44

`ConnectionManager#onLoginSuccess` is called when the `GAME_PROFILE` packet is sent to the client:
https://github.com/ViaVersion/ViaVersion/blob/master/common/src/main/java/com/viaversion/viaversion/protocols/base/BaseProtocol1_7.java#L143

However, in the bukkit main thread, what we have is code that sends packets for `PacketPlayOutLogin` (the `GAME_PROFILE` packet), as well as difficulty, location, player abilities, inventory, scoreboard, and calls PlayerJoinEvent, all in the same method and pretty close in timing to eachother

The order is roughly:
 - send `PacketLoginOutSuccess` (the `GAME_PROFILE` packet in via)
 - launch PlayerInitialSpawnEvent
 - spawn entity player in the world
 - launch PlayerSpawnLocationEvent
 - update player gamemode
 - create new PlayerConnection
 - send setup packets (login, custom payload, difficulty, spawn location, abilities, held item, entity status)
 - launch PlayerJoinEvent

This list of items is all executed on the main thread, while the packets being sent are queued on the network thread, meaning if the network thread is stressed enough, you can reach the PlayerJoinEvent and not have the player version available yet.

It is probably easier to test/debug this in PlayerInitialSpawnEvent since it occurs sooner and the race condition is probably easier to spot.

### Steps to Reproduce

[...]

PR #2849, opened 2022-03-18
Fix user connection race condition (PoC)

Pablete1234:
This is a PoC implementation of a fix for #2835

The idea is that on join event (on the main thread) we set the UserConnection in the ConnectionManager, this means any other plugin requesting player info (like the user's protocol) in PlayerJoinEvent (anything above lowest priority) will get correct info. For more info on the issue refer to #2835 .

This change makes onLoginSuccess be indepotent (assumed to be called twice for each player, once on login event from main thread, and once from network thread), and adds a listener for PlayerLoginEvent on lowest priority that finds the UserConnection from the Player -> EntityPlayer -> PlayerConnection -> NetworkManager -> Channel -> Viaversion encoder -> UserConnection

Edit: 
I have tested and validated that this works as intended on sportpaper 1.8.8 (fork of paper 1.8), i'd assume it should work on any 1.8 bukkit-based platform, if higher mc versions made changes it may not work on them, i did not test them.
 
Initially the PR had the logic on Login event, had to move it to Join event because bukkit does not create (and set) the PlayerConnection until after login event

Edit 2:
Upgraded reflection to be more robust, works in spigot 1.8.8, paper 1.8.8, sportpaper 1.8.8, paper 1.12.2, paper 1.14.4, paper 1.17.1 and paper 1.18.1, in all of the cases connecting with a 1.18.1 client

### DIFF
--- a/bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
+++ b/bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
@@ -30,42 +30,57 @@ import org.bukkit.event.Listener;
 import org.bukkit.event.player.PlayerJoinEvent;
 
 import java.lang.reflect.Field;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.logging.Level;
 
 public class JoinListener implements Listener {
 
-    private final Method getHandle;
-    private final Field connection;
-    private final Field networkManager;
-    private final Field channel;
+    private static final Method getHandle;
+    private static final Field connection;
+    private static final Field networkManager;
+    private static final Field channel;
+    private static final boolean enabled;
 
-    public JoinListener() {
+    static {
+        Method gh;
+        Field conn, nm, ch;
+        boolean en = true;
         try {
-            getHandle = NMSUtil.obc("entity.CraftPlayer").getDeclaredMethod("getHandle");
-        } catch (NoSuchMethodException | ClassNotFoundException e) {
-            throw new RuntimeException("Couldn't find CraftPlayer", e);
+            gh = NMSUtil.obc("entity.CraftPlayer").getDeclaredMethod("getHandle");
+            conn = findField(gh.getReturnType(), "PlayerConnection");
+            nm = findField(conn.getType(), "NetworkManager");
+            ch = findField(nm.getType(), "Channel");
+        } catch (NoSuchMethodException | NoSuchFieldException | ClassNotFoundException e) {
+            en = false;
+            gh = null;
+            conn = nm = ch = null;
+
+            Via.getPlatform().getLogger().log(
+                    Level.WARNING,
+                    "Couldn't find reflection methods/fields to access Channel from player.\n" +
+                            "Login race condition fixer will be disabled.\n" +
+                            " Some plugins that use ViaAPI on join event may work incorrectly.", e);
         }
-        try {
-            connection = NMSUtil.nms("EntityPlayer").getDeclaredField("playerConnection");
-        } catch (NoSuchFieldException | ClassNotFoundException e) {
-            throw new RuntimeException("Couldn't find Player Connection", e);
+
+        getHandle = gh;
+        connection = conn;
+        networkManager = nm;
+        channel = ch;
+        enabled = en;
+    }
+
+    // Loosely search a field with any name, as long as it matches the type
+    private static Field findField(Class<?> cl, String type) throws NoSuchFieldException {
+        for (Field field : cl.getDeclaredFields()) {
+            if (field.getType().getSimpleName().equals(type))
+                return field;
         }
-        try {
-            networkManager = NMSUtil.nms("PlayerConnection").getDeclaredField("networkManager");
-        } catch (NoSuchFieldException | ClassNotFoundException e) {
-            throw new RuntimeException("Couldn't find Network Manager", e);
-        }
-        try {
-            channel = NMSUtil.nms("NetworkManager").getDeclaredField("channel");
-        } catch (NoSuchFieldException | ClassNotFoundException e) {
-            throw new RuntimeException("Couldn't find Channel", e);
-        }
+        throw new NoSuchFieldException(type);
     }
 
     @EventHandler(priority = EventPriority.LOWEST)
     public void onJoin(PlayerJoinEvent e) {
+        if (!enabled) return;
         Player player = e.getPlayer();
 
         UserConnection user = getUserConnection(player);
@@ -84,9 +99,10 @@ public class JoinListener implements Listener {
 
     private UserConnection getUserConnection(Player player) {
         Channel channel = getChannel(player);
-        if (channel == null) return null;
-        BukkitEncodeHandler encoder = channel.pipeline().get(BukkitEncodeHandler.class);
-        return encoder != null ? encoder.getInfo() : null;
+        BukkitEncodeHandler encoder;
+        if (channel != null && (encoder = channel.pipeline().get(BukkitEncodeHandler.class)) != null)
+            return encoder.getInfo();
+        return null;
     }
 
     private Channel getChannel(Player player) {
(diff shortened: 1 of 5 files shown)


bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        ProtocolInfo info = user.getProtocolInfo();
+        info.setUuid(player.getUniqueId());
+        info.setUsername(player.getName());
+        Via.getManager().getConnectionManager().onLoginSuccess(user);
+    }
+
+    private UserConnection getUserConnection(Player player) {
+        Channel channel = getChannel(player);
```
kennytv, line 101: Check if the channel is still open

bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
```diff
@@ -0,0 +1,120 @@
...
+        BukkitEncodeHandler encoder;
+        if (channel != null && (encoder = channel.pipeline().get(BukkitEncodeHandler.class)) != null)
```
kennytv, line 103: Opening/closing brackets

bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
```diff
@@ -0,0 +1,120 @@
...
+        for (Field field : cl.getDeclaredFields()) {
+            if (field.getType().getSimpleName().equals(type))
+                return field;
+        }
+        throw new NoSuchFieldException(type);
+    }
+
+    @EventHandler(priority = EventPriority.LOWEST)
+    public void onJoin(PlayerJoinEvent e) {
+        if (!enabled) return;
```
kennytv, line 83: Just check if one of the fields is null

bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
```diff
@@ -0,0 +1,120 @@
...
+
+    private static final Method getHandle;
+    private static final Field connection;
+    private static final Field networkManager;
+    private static final Field channel;
+    private static final boolean enabled;
+
+    static {
+        Method gh;
+        Field conn, nm, ch;
```
kennytv, line 46: Initialize with null and remove those in the catch block

bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
```diff
@@ -0,0 +1,120 @@
...
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.logging.Level;
+
+public class JoinListener implements Listener {
+
+    private static final Method getHandle;
+    private static final Field connection;
+    private static final Field networkManager;
+    private static final Field channel;
```
kennytv, line 41: upper case names

kennytv (changes requested):

bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
Pablete1234: then you have a situation where some may be set and others not set, depending on what threw the exception, instead of the current consistency of enabled -> all non-null, not-enabled -> all null

bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
Pablete1234: what thread does the channel close in, wouldn't that be grounds for a new possible race condition?

User logs in, and in network thread, connection closes. Later the PlayerJoinEvent (on main thread) re-sets the same player.
If channel closing is in fact in network thread, checking here for open only reduces how likely it is that we find that race condition, but we may still get such a case

kennytv: It can be closed from anywhere, but we have a close future listener that includes already closed channels, so that doesn't matter

Pablete1234:
I have synchronized onLoginSuccess, without making that method atomic, you could have the following order of events:

1 - network thread calls onLoginSuccess fully
2 - main thread calls onLoginSuccess, newlyAdded = false (it wasn't just added)
3 - the user disconnects, the listener calls onDisconnect
4 - the main thread continues the onLoginSuccess execution, re-adding to the map of uuid-connection, and without adding a listener to cleanup later

common/src/main/java/com/viaversion/viaversion/connection/ConnectionManagerImpl.java
```diff
@@ -35,24 +35,25 @@
     protected final Set<UserConnection> connections = Collections.newSetFromMap(new ConcurrentHashMap<>());
 
     @Override
-    public void onLoginSuccess(UserConnection connection) {
+    public synchronized void onLoginSuccess(UserConnection connection) {
```
kennytv, line 38: Don't synchronize, use an atomic boolean on UserConnection for having been added that's compared and set at the beginning of this method to return early

common/src/main/java/com/viaversion/viaversion/connection/ConnectionManagerImpl.java
```diff
@@ -35,24 +35,25 @@
...
 
-        if (connection.getChannel() != null) {
+        if (newlyAdded && connection.getChannel() != null) {
             connection.getChannel().closeFuture().addListener((ChannelFutureListener) future -> onDisconnect(connection));
         }
     }
 
     @Override
-    public void onDisconnect(UserConnection connection) {
+    public synchronized void onDisconnect(UserConnection connection) {
```
kennytv, line 56: No need to synchronize the onDisconnect at all

bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
Pablete1234: yeah, but nothing prevents us re-adding (by calling onLoginSuccess) *after* it had already closed, leaking the UserConnection, this could happen if main thread is laggy enough that on network thread you fully joined and disconnected, and you hit the timings *just right* in a very obscure way

kennytv (changes requested):

bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
kennytv: No, the close future is executed when set and the channel is already closes, immediately removing it again. but that'd be alleviated by the latest review in any case

common/src/main/java/com/viaversion/viaversion/connection/ConnectionManagerImpl.java
Pablete1234: Without synchronizing onDisconnect, onLoginSuccess can be adding at-the-same-time as disconnect is removing, where onLoginSuccess tries to connections.add which returns false, as in, connection is already there, and onDisconnect removes it instants later

Then onLoginSuccess proceeds to add it to the map, (where it will be left forever) and a closeFuture will not be dispatched, as "newlyAdded" was false

common/src/main/java/com/viaversion/viaversion/connection/ConnectionManagerImpl.java
Pablete1234: that approach would not fix the race condition

The network thread may start handling the login success, setting that atomic boolean to true
The main thread reads the boolean, sees it's already handled, and early-returns
The next plugin ties to access ViaAPI, and the player may, or may not, be set, since it's the network thread handling the adding-to-the-map

kennytv:
Looks better 👍 I'll properly review in a few days or so

calcastor:
Looking forward to this being further reviewed and/or merged; we've had this PR deployed in a production environment for a few months now with no issues. It has solved the relevant issue in our experience.

kennytv:
Sorry, this completely went under my radar. I'll make sure to review it again soon-ish

kennytv (commented):
I also think I'll have to take back what I said on the synchronized methods

Pablete1234:
Rebased ontop of master & added support for mojmap names, as well as addressing the other concerns on the review

kennytv (approved):

Mercenarioo:
The issue is still there https://github.com/ViaVersion/ViaVersion/issues/3140

Issue #2915, opened 2022-06-07
ViaVersion 4.3.0 crashes Velocity with 1.19 client


seb303:
### `/viaversion dump` Output

https://dump.viaversion.com/1dd847c60a3bfaa7d8045e74ea519f80ed2a01bee8fa1c3cee687ea89ad91c1d

### Console Error

```
Error on join: https://paste.gg/p/anonymous/fd51f19d66204f1290f4f26fb24c4605
Joined ok, error on change world: https://paste.gg/p/anonymous/e6fff895ef624db9940cb3fdff6ac3ed
```

### Bug Description

Sometimes Velocity crashes immediately when a player joins.
Always it happens when changing worlds.
There are no custom dimensions used on the server, despite the error message.  Just the usual 3.

### Steps to Reproduce

1. Install Velocity 3.1.2-SNAPSHOT with ViaVersion 4.3.0 & ViaBackwards 4.3.0
2. Link to 1.17.1 Paper backend
3. Connect with 1.19 client (the problem does not occur with 1.18.2 or 1.17.1 clients)
5. If the error doesn't appear immediately, try changing worlds.

### Expected Behavior

Crash should not happen (the problem does not occur with 1.18.2 or 1.17.1 clients).

### Additional Server Info

Proxy is Velocity 3.1.2-SNAPSHOT
The following plugins were also installed on the proxy when the logs were recorded: Essential-Playerinfo-Plugin-2.0.1, Geyser-Velocity, floodgate-velocity (both latest snapshot), ViaRewind-2.0.3-SNAPSHOT
Note however the the problem happens just the same with all the above plugins removed.

Backend server is 1.17.1 Paper, with Plugins: AntiCooldown, BuildBattle, Citizens, CoreProtect, Denizen, Depenizen, Duels, ExtraSpecialItems, FireBallKnockback, Gen-Splitter, HubBasics, Images, KeepChunks, LuckPerms, MBedwars, MBedwarsTweaks, Multiverse-Core, Multiverse-Inventories, Multiverse-NetherPortals, Multiverse-Portals, Parkour, PlaceholderAPI, ProtocolLib, SimpleTpa*, Skript, Skywars, Skywars-Extension*, sleep-most, SlimeWorldManager, TAB, TheBridge, TubeTils, WorldEdit, WorldGuard

### Checklist

[...]

PR #2990, opened 2022-06-29
Use most similar dimension when it's not found

creeper123123321:
workaround for ViaVersion/ViaFabric#182 and #2915


kennytv:
Should at least make sure the height values are the same (`min_y`, `height`), then go by difference

kennytv (approved):

PR #3079, opened 2022-08-04
add flamecord warning message

tomcraft:
Since Flamecord doesn't work with ViaVersion, I think it could be great to add it as an unsupported software to log something users can understand without getting to Discord for support.

I also updated the protocol version constant check for bungeecord outdated check.


common/src/main/java/com/viaversion/viaversion/unsupported/UnsupportedServerSoftware.java
```diff
@@ -110,5 +110,6 @@ public UnsupportedSoftware build() {
     public static final class Reason {
 
         public static final String DANGEROUS_SERVER_SOFTWARE = "You are using server software that - outside of possibly breaking ViaVersion - can also cause severe damage to your server's integrity as a whole.";
+        public static final String BREAKING_PROXY_SOFTWARE = "You are using proxy software that intentionally breaks ViaVersion. Please use another proxy software or move ViaVersion to each servers instead of proxy.";
```
kennytv, line 113: to each backend server instead of the proxy

tomcraft:
made requested changes

kennytv (approved):

PR #3126, opened 2022-09-18
Preparation for remapping the 1.19.0 signatures in ViaFabric

FlorianMichael:
This pull request is a preparation for fixing the signatures of the 1.19.2 -> 1.19.0 protocol, it removes the old code from ViaVersion and just lets the input from the original client pass through, so that you can later implement the signatures of 1.19.0 in ViaFabric so that you can join. If this PR is merged, I will make a second one in ViaFabric that uses the provider.


kennytv:
It'd be easier for ViaFabric to just replace those handlers if it needs to

FlorianMichael:
So I should rather modify this part of ViaVersion in my ViaFabric PR?

kennytv:
Yes, just register packet remappers with the override flag set on those on the right protocol if the option is enabled in VF

FlorianMichael:
All right, then I will do it like this

FlorianMichael:
ViaFabric PR is done, https://github.com/ViaVersion/ViaFabric/pull/200

PR #3127, opened 2022-09-19
fixed Particle Rendering in 1.13.0 -> 1.12.2

FlorianMichael:
As of Minecraft 1.13, a new flag has been added to the Status Effect Packet that specifically says whether a Status Effect should be displayed in-game at the top right, this PR always sets it to true if enabled in ViaConfig. So that 1.14.4+ clients on older servers have the HUD correctly.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/EntityPackets.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/EntityPackets.java
@@ -159,7 +159,7 @@ public class EntityPackets {
                     byte flags = packetWrapper.read(Type.BYTE); // Input Flags
 
                     if (Via.getConfig().isNewEffectIndicator())
-                        flags += 2; // Since Minecraft 1.14.4, Minecraft has 2 Flags for rendering the particles, one for the Player Inventory, and one for the InGame HUD
+                        flags += 2;
 
                     packetWrapper.write(Type.BYTE, flags);
                 });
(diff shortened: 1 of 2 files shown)


Sumandora (approved):
Changed my life, fantastic!

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/EntityPackets.java
```diff
@@ -146,6 +147,25 @@ public void registerMap() {
...
+                map(Type.VAR_INT); // Entity id
+                map(Type.BYTE); // Effect id
+                map(Type.BYTE); // Amplifier
+                map(Type.VAR_INT); // Duration
+
+                handler(packetWrapper -> {
+                    byte flags = packetWrapper.read(Type.BYTE); // Input Flags
+
+                    if (Via.getConfig().isNewEffectIndicator())
+                        flags += 2;
```
tomcraft, line 162: Looking at wiki.vg it looks like a bitmask so, I think this should be `flags |= 0x02` to make sure this does not shift anything else.

FlorianMichael: |= 0x02 does not work, besides there is no other use case anyway, have you tried if your approach works?

FlorianMichael: fixed

tomcraft:
Ok this seems good to me, I had a hard time figuring out what was this about because this flag is not referenced in the wiki.vg page, after looking at 1.13 source code I can say this is the good usage of the flag.

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/EntityPackets.java
tomcraft: After looking at source code it looks like 0x04 is the right mask to apply 👍

kennytv (approved):

PR #3129, opened 2022-09-28
add sanity check for some block nbt

creeper123123321:


common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/InventoryPackets.java
```diff
@@ -309,8 +309,12 @@ public Item handleItemToClient(Item item) {
                     if (blockEntityTag.get("Patterns") instanceof ListTag) {
                         for (Tag pattern : (ListTag) blockEntityTag.get("Patterns")) {
                             if (pattern instanceof CompoundTag) {
-                                IntTag c = ((CompoundTag) pattern).get("Color");
-                                c.setValue(15 - c.asInt()); // Invert color id
+                                Tag c = ((CompoundTag) pattern).get("Color");
+                                if (c instanceof NumberTag) {
+                                    // Invert color id
+                                    ((CompoundTag) pattern).put("Color",
+                                            new IntTag(15 - ((IntTag) c).asInt()));
```
tomcraft, line 316: You check for NumberTag but then you cast to IntTag.

tomcraft (commented):
I think it adds great safety against bad custom implementations

kennytv (approved):

PR #3134, opened 2022-10-04
Some minor performance improvements and cleanups

Gerrygames:
Replaced some x,y,z for loops with index loops to reduce xyz/index packing/unpacking. Also replaced calls to methods that access the palette via ChunkSection with direct calls to the palette. And some other minor changes to reduce object allocation.
Did some quick testing with a 1.19.2 client on a 1.8.9 server, but further testing might be needed.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/WorldPackets.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/WorldPackets.java
@@ -27,6 +27,8 @@ import com.viaversion.viaversion.api.minecraft.BlockChangeRecord;
 import com.viaversion.viaversion.api.minecraft.Position;
 import com.viaversion.viaversion.api.minecraft.chunks.Chunk;
 import com.viaversion.viaversion.api.minecraft.chunks.ChunkSection;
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/WorldPackets.java
```diff
@@ -334,77 +336,63 @@ public void registerMap() {
...
-                                            ConnectionData.blockConnectionProvider.storeBlock(wrapper.user(), globalX,
-                                                    globalY, globalZ, block);
-                                        }
-                                    }
+                            for (int idx = 0; idx < ChunkSection.SIZE; idx++) {
+                                int id = blocks.idAt(idx);
+                                if (ConnectionData.isWelcome(id)) {
+                                    int globalX = (idx & 0xF) + (chunk.getX() << 4);
+                                    int globalY = (idx >> 8 & 0xF) + (s << 4);
+                                    int globalZ = (idx >> 4 & 0xF) + (chunk.getZ() << 4);
```
kennytv, line 394: Should probably add static `xFromIndex`, y, z methods in ChunkSection so these kinds of lines are easier to understand

kennytv (approved):
Also remove the assertions, else looks good

PR #3138, opened 2022-10-12
Fixes an issue with the 0x00 packet and how it's handled with ViaVers…

Summiner:
Fixes an issue with the 0x00 packet and how it's handled with ViaVersion - Leads to servers crashing & more
Example: https://www.youtube.com/watch?v=AB6OB0W_3CU


bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/JoinListener.java
```diff
@@ -23,6 +23,7 @@
 import com.viaversion.viaversion.bukkit.handlers.BukkitEncodeHandler;
 import com.viaversion.viaversion.bukkit.util.NMSUtil;
 import io.netty.channel.Channel;
+import jdk.javadoc.internal.doclets.formats.html.markup.Text;
```
Camotoy, line 26: Unused import.

Camotoy:
Has this actually been tested? The error is probably thrown here: https://github.com/ViaVersion/ViaVersion/blob/master/common/src/main/java/com/viaversion/viaversion/protocols/base/BaseProtocol1_7.java#L140 - the solution presented here would be after the string is decoded.

PR #3147, opened 2022-10-26
Fix missing zombie_pigman (1.15.2) => zombified_piglin (1.16) translation in mob_spawner tile entity

valaphee:
# Description
Between version 1.15.2 and 1.16, the entity zombie_pigman was renamed to zombified_piglin, but mob_spawners didn't had any translation, which causes the client to spam the log with errors "Skipping Entity with id minecraft:zombie_pigman" and no entity is shown in the mob spawner.

# Testing
Can be tested with a server which has version 1.8.9 until 1.15.2 and a client with version 1.16 until 1.19.2 by creating a mob spawner, and setting it with a zombified piglin spawnegg.


Kichura (approved):
Nice catch.

kennytv (approved):
Thanks!

PR #3148, opened 2022-10-30
Emulates Armor HUD for 1.9+ players on 1.8 servers

FlorianMichael:
This pull request emulates the Armor of each 1.9+ player and sends back the correct Entity Properties packet, so that the Armor HUD of Minecraft for 1.9+ players works again on 1.8 servers.

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
@@ -396,6 +396,13 @@ public interface ViaVersionConfig {
     boolean isLeftHandedHandling();
 
     /**
+     * Emulates the armor hud on 1.8 servers for 1.9+ clients
+     *
+     * @return true if enabled
+     */
+    boolean isArmorHud1_8();
+
+    /**
      * Fixes velocity bugs due to different hitbox for 1.9-1.13 clients on 1.8 servers.
      *
      * @return true if enabled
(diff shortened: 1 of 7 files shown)


tomcraft:
I think there was already something to deal with this. Maybe it was a bukkit impl only thing.

joao218623:
I can't make bridges with viarewind i disable everything the bossbar... but i can't it makes like the block doens't exit

FlorianMichael:
> I think there was already something to deal with this. Maybe it was a bukkit impl only thing.

It's also a fix for ViaFabric

FlorianMichael:
> I can't make bridges with viarewind i disable everything the bossbar... but i can't it makes like the block doens't exit

i think you are on the wrong issue thread?

tomcraft:
> > I think there was already something to deal with this. Maybe it was a bukkit impl only thing.
> 
> It's also a fix for ViaFabric

Then if it works it's fine, but you should remove the previous code in the bukkit listener. 👍

FlorianMichael:
> > > I think there was already something to deal with this. Maybe it was a bukkit impl only thing.
> > 
> > 
> > It's also a fix for ViaFabric
> 
> Then if it works it's fine, but you should remove the previous code in the bukkit listener. 👍

The bukkit code is much better than the protocol code because it has access to the player inventory

FlorianMichael:
I think then it would be better to fix it only in ViaFabric, I didn't know that ViaVersion for Bukkit already has a fix on platform level, I'll just wait for Kenny's opinion

kennytv:
Generally having it on a plain protocol level (as well) is nice, but since it's 1.8 and touching the inventory, I agree it'd probably just be better in ViaFabric to have the two most important platforms covered in a better way than would be possible on a pure protocol level

FlorianMichael:
Yes ok, I will make a ViaFabric pull request tomorrow

PR #3156, opened 2022-11-19
Fix recipe book not being empty when using a multi-server configuration, with servers which are pre-1.12.

valaphee:
# Description
When using a multi-server configuration, with servers which are pre-1.12, switching from 1.12 to pre-1.12 causes the recipe book to not clear.

# Testing
Can be tested with a BungeeCord and two servers, one is pre-1.12 and the other one 1.12 or later. When switching from 1.12 to 1.11, the recipe book is still shown, even when unusable and containing invalid recipes.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_12to1_11_1/Protocol1_12To1_11_1.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_12to1_11_1/Protocol1_12To1_11_1.java
@@ -33,6 +33,7 @@ import com.viaversion.viaversion.api.protocol.AbstractProtocol;
 import com.viaversion.viaversion.api.protocol.packet.PacketWrapper;
 import com.viaversion.viaversion.api.protocol.remapper.PacketHandler;
 import com.viaversion.viaversion.api.protocol.remapper.PacketRemapper;
+import com.viaversion.viaversion.api.protocol.version.ProtocolVersion;
 import com.viaversion.viaversion.api.rewriter.ItemRewriter;
 import com.viaversion.viaversion.api.type.Type;
 import com.viaversion.viaversion.api.type.types.version.Types1_12;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/protocol1_12to1_11_1/Protocol1_12To1_11_1.java
```diff
@@ -167,6 +169,13 @@ public void registerMap() {
                     ClientWorld clientChunks = wrapper.user().get(ClientWorld.class);
                     int dimensionId = wrapper.get(Type.INT, 1);
                     clientChunks.setEnvironment(dimensionId);
+
+                    wrapper.create(ClientboundPackets1_13.DECLARE_RECIPES, new PacketHandler() { // Reset recipes
+                        @Override
+                        public void handle(PacketWrapper wrapper) throws Exception {
+                            wrapper.write(Type.VAR_INT, 0);
+                        }
+                    }).scheduleSend(Protocol1_13To1_12_2.class);
```
kennytv, line 178: Please make the handler a lambda

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_12to1_11_1/Protocol1_12To1_11_1.java
```diff
@@ -167,6 +169,10 @@ public void registerMap() {
                     ClientWorld clientChunks = wrapper.user().get(ClientWorld.class);
                     int dimensionId = wrapper.get(Type.INT, 1);
                     clientChunks.setEnvironment(dimensionId);
+
+                    // Reset recipes
+                    wrapper.create(ClientboundPackets1_13.DECLARE_RECIPES, packetWrapper -> packetWrapper.write(Type.VAR_INT, 0))
+                            .scheduleSend(Protocol1_13To1_12_2.class);
```
kennytv, line 175: Oh and this should be `Protocol1_12To1_11_1.class`, the default behavior of that method is to skip current pipeline (aka start from the target)

kennytv:
What I missed: You need to make sure the client is a actually on 1.13+ before sending something through the 1.13 pipeline

valaphee:
Ah thanks, wasn't sure if it just gets omitted

kennytv (approved):
Thanks!

PR #3210, opened 2023-02-02
Add armor toggle patch for Bukkit

Camotoy:
1.19.4 will allow players to swap armor as long as the armor slot isn't identical to the hand slot. This PR patches that using the Bukkit API.


bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/protocol1_19_4To1_19_3/ArmorToggleListener.java
```diff
@@ -0,0 +1,55 @@
...
+import org.bukkit.inventory.EquipmentSlot;
+import org.bukkit.inventory.ItemStack;
+
+public final class ArmorToggleListener extends ViaBukkitListener {
+
+    public ArmorToggleListener(ViaVersionPlugin plugin) {
+        super(plugin, Protocol1_19_4To1_19_3.class);
+    }
+
+    @EventHandler
```
kennytv, line 36: Should be monitor and ignore cancelled

kennytv (approved):

PR #3221, opened 2023-02-23
fix #2912

### LINKED ISSUE #2912: Wrong version range for protocol 4
### Problem Description

only 1.7.2-1.7.5 's protocol is 4 and supported.
1.7(.0)-1.7.1's protocol is 3 and it's not supported

### Solution Description

ProtocolVersion.java:50
public static final ProtocolVersion v1_7_2 = register(4, "1.7.2-1.7.5", new VersionRange("1.7", 2, 5));

### Alternatives

_No response_

### Additional Info

little but correct

HowardZHY:
v1_7_1 didn't rename for compatibility


kennytv (changes requested):
We don't track snapshot versions. 1.7 and 1.7.1 don't exist

kennytv (approved):

PR #3228, opened 2023-03-04
Optimize block connections on neighbour chunk calculation

Pablete1234:
Optimizes the following bottleneck:
image

This is a profile from the netty event thread, for a few seconds during which a ton of players (about 60) are teleported between worlds. Updating block connections for neighboring chunks was a pretty significant part of the time, causing some lag on the network thread and causing players to run behind the server for a bit.

The PR mainly caches `connection.get(BlockConnectionStorage.class)` and `blockStorage.get(index)`, both of which are taking significantly long and don't need to be ran *for every single block* when the value returned can be cached temporarily.

Note: I have not yet ran/tested these changes, but i'm still opening the PR as a draft to be able to get feedback/reviews. Will update with results once i've tested, deployed, and validated the results.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/blockconnections/ConnectionData.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/blockconnections/ConnectionData.java
@@ -18,7 +18,6 @@
 package com.viaversion.viaversion.protocols.protocol1_13to1_12_2.blockconnections;
 
 import com.github.steveice10.opennbt.tag.builtin.ListTag;
-import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.viaversion.viaversion.api.Via;
@@ -37,6 +36,7 @@ import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.ClientboundPacke
 import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.Protocol1_13To1_12_2;
 import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.blockconnections.providers.BlockConnectionProvider;
 import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.blockconnections.providers.PacketBlockConnectionProvider;
+import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.blockconnections.providers.UserBlockData;
 import com.viaversion.viaversion.util.Key;
 import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
 import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/blockconnections/ConnectionData.java
```diff
@@ -693,4 +619,89 @@ interface ConnectorInitAction {
...
+                        }
+                    }
+
+                    if (!updates.isEmpty()) {
+                        PacketWrapper wrapper = PacketWrapper.create(ClientboundPackets1_13.MULTI_BLOCK_CHANGE, null, user);
+                        wrapper.write(Type.INT, chunkX + chunkDeltaX);
+                        wrapper.write(Type.INT, chunkZ + chunkDeltaZ);
+                        wrapper.write(Type.BLOCK_CHANGE_RECORD_ARRAY, updates.toArray(EMPTY_RECORDS));
+                        try {
+                            wrapper.send(Protocol1_13To1_12_2.class);
```
kennytv, line 687: Although from the previous code, please make the method throw Exception and remove the try catch here

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/storage/BlockConnectionStorage.java
```diff
@@ -85,7 +89,7 @@ public int get(int x, int y, int z) {
 
     public void remove(int x, int y, int z) {
```
kennytv, line 90: Also need to set the cached object to null at the end of the this method where a chunk entry is removed if empty

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/storage/BlockConnectionStorage.java
```diff
@@ -118,11 +123,15 @@ public void unloadChunk(int x, int z) {
...
+            lastSection = null;
+        }
     }
 
     private SectionData getChunkSection(long index, boolean requireNibbleArray) {
-        SectionData map = blockStorage.get(index);
+        SectionData map = getSection(index);
         if (map == null) {
             map = new SectionData(new byte[4096]);
             blockStorage.put(index, map);
```
kennytv, line 137: Could probably also set the cached value here, assuming it's going to be called multiple times after computed

Pablete1234:
We've been running this during the day, it's been running smoothly (stability-wise), and it does seem like performance has gotten better.
image

The new profile shows that calls to `connection.get(BlockConnectionStorage.class)` are indeed gone, however the other one doesn't show as much of an improvement as i expected. Unsure if getSection is most of the times missing the cache, or if fastutil was already smart about retrieving the same section as before. Regardless, the changes are positive for performance.

Pablete1234:
It seems like support for caching the `null` chunk section has a big impact, and my assumption is that most of the times when loading a chunk the chunk next to it isn't there yet, in which case the cache optimization was doing nothing.
After those changes, the effect of getBlockData (in comparison to updateBlock) has become negligible:
image
0.14/3.99 = roughly 3.5% vs prior to all these changes it being 13.92/23.84, roughly 58.3%

kennytv (approved):
Looks good, thanks!

PR #3267, opened 2023-03-18
Patch armor swapping on 1.19.3 servers

Tisawesomeness:
Updates the armor toggle listener to swap armor rather than update to original inventory, but only when action is `RIGHT_CLICK_AIR` to prevent double toggle when `LEFT_CLICK_AIR` event also fires

### DIFF
--- a/bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/protocol1_19_4To1_19_3/ArmorToggleListener.java
+++ b/bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/protocol1_19_4To1_19_3/ArmorToggleListener.java
@@ -40,7 +40,10 @@ public final class ArmorToggleListener extends ViaBukkitListener {
     public void itemUse(final PlayerInteractEvent event) {
         final Player player = event.getPlayer();
         final ItemStack item = event.getItem();
-        if (item == null || event.getHand() == null || event.getAction() != Action.RIGHT_CLICK_AIR || !isOnPipe(player)) {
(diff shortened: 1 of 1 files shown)


kennytv:
We don't add features from newer client or server versions (especially if kind of intrusive like this), the listener exists to make sure no desync happens. The change to the return condition is fine though

bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/protocol1_19_4To1_19_3/ArmorToggleListener.java
```diff
@@ -39,7 +40,7 @@ public ArmorToggleListener(final ViaVersionPlugin plugin) {
     public void itemUse(final PlayerInteractEvent event) {
         final Player player = event.getPlayer();
         final ItemStack item = event.getItem();
-        if (item == null || event.getHand() == null || !isOnPipe(player)) {
+        if (item == null || event.getHand() == null || event.getAction() != Action.RIGHT_CLICK_AIR || !isOnPipe(player)) {
```
kennytv, line 43: also needs right click block

Tisawesomeness:
Fair, though some 1.19.4 players may expect armor swapping to work and see the inventory resetting as a bug. Maybe a config toggle for armor swapping behavior (default no swapping since it's an intrusive change)? If that's out of scope for ViaVersion, I can release a separate plugin for those who want it.

tomcraft:
Yes a separate plugin is more appropriate for this kind of feature as it is not tied to protocol translation. It could work on any version, not specially 1.19.3.

PR #3342, opened 2023-06-08
Rotate item displays by 180 degrees around the y-axis

MrKinau:
Items in item display entities got rotated in 1.20. This PR reverses this behaviour for 1.20 players to keep the rotation consistent for all players.

> `item_display` items have been rotated 180 degrees around the Y axis to better match the transformation that is applied when rendering items on an Armor Stand head and in Item Frames

Not sure if I added the handlers correctly, because I needed to re-map all packets containing the entity rotation to add the handlers manipulating the yaw and pitch of the entities, maybe there is another way to do so without re-mapping them.


kennytv:
Thank, manual merge in https://github.com/ViaVersion/ViaVersion/commit/2a14f9a0cdc8676ee4ebada268bf57ce48d17ab8

PR #3359, opened 2023-06-18
Fix missing block connection updates on cancelled place/remove

Pablete1234:
Currently when using the packet-based block storage for block connection resolving, placing a block (or breaking it) and the event being cancelled by the server, leaves fences and similar in an invalid/broken state.

Example:
broken-block-connections

The reason this is broken is that the following sequence of events occurs:
 1) (prior) Load the world/chunk, all the block-states are saved per player based on the packets sent to them and computed block connections.
 2) Place (or break) the block
 3) Client-side, your block gets connected to the newly placed block
 4) Server-side, you're sent an update that the block should be air (your place was cancelled)
 5) Viaversion tries to update neighbor connections, but sees the fence as not-connected, as saved in step 1, and so skips the update.

There's not an unique solution to fixing this, the simplest would be to simply remove the whole `if (newBlockState == blockState) ... continue` block, and always send the updates, but it's not too efficient to be sending all updates to all players when most do not need the update.

The approach in this PR instead saves whenever a user places/breaks a block (stores the last 5 place/break) and if an update occurs near any of those it will be forcibly sent, and won't be optimized away. The reason for it being "a close block" is that the block-place packet does not guarantee where the block will place; the position is the block you're placing *against*, and sometimes the block places next to it, while others it places inside it (eg: vines or 1-thick snow), so i just avoided the logic of that and simply did a taxicab distance < 2, which means the referenced block or any direct neighbor is valid for forced re-sending.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/blockconnections/ConnectionData.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/blockconnections/ConnectionData.java
@@ -76,9 +76,13 @@ public final class ConnectionData {
 
             int newBlockState = handler.connect(user, pos, blockState);
             if (newBlockState == blockState) {
-                if (inSync == null) inSync = blockConnectionProvider.storesBlocks(user, position);
+                if (inSync == null) {
+                    inSync = blockConnectionProvider.storesBlocks(user, position);
+                }
                 // Blocks-states are the same, and known to be stored and not de-synced, skip update
-                if (inSync) continue;
+                if (inSync) {
+                    continue;
+                }
             }
 
             updateBlockStorage(user, pos.x(), pos.y(), pos.z(), newBlockState);
@@ -735,7 +739,7 @@ public final class ConnectionData {
 
             Position pos = new Position(x, y, z);
             int newBlockState = handler.connect(user, pos, blockState);
-            if (blockState != newBlockState || !blockConnectionProvider.storesBlocks(user, pos)) {
+            if (blockState != newBlockState || !blockConnectionProvider.storesBlocks(user, null)) {
                 records.add(new BlockChangeRecord1_8(x & 0xF, y, z & 0xF, newBlockState));
                 updateBlockStorage(user, x, y, z, newBlockState);
             }
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/blockconnections/ConnectionData.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/blockconnections/ConnectionData.java
@@ -64,6 +64,8 @@ public final class ConnectionData {
     }
 
     public static void update(UserConnection user, Position position) throws Exception {
+        Boolean inSync = null;
+
         for (BlockFace face : BlockFace.values()) {
             Position pos = position.getRelative(face);
             int blockState = blockConnectionProvider.getBlockData(user, pos.x(), pos.y(), pos.z());
@@ -73,8 +75,10 @@ public final class ConnectionData {
             }
 
             int newBlockState = handler.connect(user, pos, blockState);
-            if (newBlockState == blockState && blockConnectionProvider.storesBlocks()) {
-                continue;
+            if (newBlockState == blockState) {
+                if (inSync == null) inSync = blockConnectionProvider.storesBlocks(user, position);
+                // Blocks-states are the same, and known to be stored and not de-synced, skip update
+                if (inSync) continue;
             }
 
             updateBlockStorage(user, pos.x(), pos.y(), pos.z(), newBlockState);
@@ -100,8 +104,13 @@ public final class ConnectionData {
         blockConnectionProvider.clearStorage(connection);
     }
 
+    public static void markModified(UserConnection connection, Position pos) {
+        if (!needStoreBlocks()) return;
+        blockConnectionProvider.modifiedBlock(connection, pos);
+    }
+
     public static boolean needStoreBlocks() {
-        return blockConnectionProvider.storesBlocks();
+        return blockConnectionProvider.storesBlocks(null, null);
     }
 
     public static void connectBlocks(UserConnection user, Chunk chunk) {
@@ -726,7 +735,7 @@ public final class ConnectionData {
 
             Position pos = new Position(x, y, z);
             int newBlockState = handler.connect(user, pos, blockState);
-            if (blockState != newBlockState || !blockConnectionProvider.storesBlocks()) {
+            if (blockState != newBlockState || !blockConnectionProvider.storesBlocks(user, pos)) {
                 records.add(new BlockChangeRecord1_8(x & 0xF, y, z & 0xF, newBlockState));
                 updateBlockStorage(user, x, y, z, newBlockState);
             }
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/WorldPackets.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/WorldPackets.java
@@ -36,6 +36,7 @@ import com.viaversion.viaversion.api.type.types.Particle;
 import com.viaversion.viaversion.protocols.protocol1_12_1to1_12.ClientboundPackets1_12_1;
 import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.ClientboundPackets1_13;
 import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.Protocol1_13To1_12_2;
+import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.ServerboundPackets1_13;
 import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.blockconnections.ConnectionData;
 import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.blockconnections.ConnectionHandler;
 import com.viaversion.viaversion.protocols.protocol1_13to1_12_2.data.NamedSoundRewriter;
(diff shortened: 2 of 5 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/WorldPackets.java
```diff
@@ -535,6 +537,32 @@ else if (particleId == 36)
...
+            }
+        });
+        protocol.registerServerbound(ServerboundPackets1_13.PLAYER_DIGGING, wrapper -> {
+            Position pos;
+            wrapper.write(Type.VAR_INT, wrapper.read(Type.VAR_INT)); // Status
+            wrapper.write(Type.POSITION, pos = wrapper.read(Type.POSITION)); // Location
+            wrapper.write(Type.UNSIGNED_BYTE, wrapper.read(Type.UNSIGNED_BYTE)); // block face
+
+            if (Via.getConfig().isServersideBlockConnections() && ConnectionData.needStoreBlocks()) {
+                ConnectionData.markModified(wrapper.user(), pos);
```
RaphiMC, line 562: You have to check the status here because PLAYER_DIGGING is abused for other actions than block breaking

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/WorldPackets.java
```diff
@@ -535,6 +537,32 @@ else if (particleId == 36)
                 });
             }
         });
+
+        // Incoming Packets
+        protocol.registerServerbound(ServerboundPackets1_13.PLAYER_BLOCK_PLACEMENT, wrapper -> {
+            Position pos;
+            wrapper.write(Type.POSITION, pos = wrapper.read(Type.POSITION)); // Location
```
RaphiMC, line 544: Consider using map() instead of write(read()) like done in the SPAWN_PARTICLE handler

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/WorldPackets.java
```diff
@@ -535,6 +537,32 @@ else if (particleId == 36)
...
+            wrapper.write(Type.FLOAT, wrapper.read(Type.FLOAT)); // cursor y
+            wrapper.write(Type.FLOAT, wrapper.read(Type.FLOAT)); // cursor z
+
+            if (Via.getConfig().isServersideBlockConnections() && ConnectionData.needStoreBlocks()) {
+                ConnectionData.markModified(wrapper.user(), pos);
+            }
+        });
+        protocol.registerServerbound(ServerboundPackets1_13.PLAYER_DIGGING, wrapper -> {
+            Position pos;
+            wrapper.write(Type.VAR_INT, wrapper.read(Type.VAR_INT)); // Status
```
RaphiMC, line 557: Consider using map() instead of write(read()) like done in the SPAWN_PARTICLE handler

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/storage/BlockConnectionStorage.java
```diff
@@ -17,18 +17,24 @@
...
+
+import com.viaversion.viaversion.api.minecraft.Position;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class BlockConnectionStorage implements StorableObject {
     private static Constructor<?> fastUtilLongObjectHashMap;
 
     private final Map<Long, SectionData> blockStorage = createLongObjectMap();
+    @SuppressWarnings("UnstableApiUsage")
+    private final Queue<Position> modified = EvictingQueue.create(5);
```
RaphiMC, line 37: This should be larger as players might place/break more than 5 blocks per whatever the ping to the server is seconds

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/storage/BlockConnectionStorage.java
```diff
@@ -88,10 +94,23 @@ public void remove(int x, int y, int z) {
         }
     }
 
+    public void markModified(Position pos) {
+        this.modified.add(pos);
+    }
+
+    public boolean recentlyModified(Position pos) {
```
RaphiMC, line 101: Using a global queue which never has less than 5 elements in it is a very big performance issue. Block connections by themself are already pretty slow, but adding at least 5 (but more likely even more) additional iterations to each block (inside a chunk) being checked/stored is way too much for solving such a small issue

RaphiMC (changes requested):

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/storage/BlockConnectionStorage.java
Pablete1234: The performance on block connections is especially terrible for connecting neighbours across chunks, but that shouldnt be using this at all.

The affected performance here will be:
 - for every modified block
 - if the block is next to a fence/whetever
 - if the fence, on the client-side, has exactly the same shape as we've resolves it should have
 - then we check a list of at most 5 blocks

So no, not each single block is getting the 5 extra iterations, only the edge-case ones

That's also the reason i wanted to keep the list small, even if it has the potential for a player to place 10 blocks within ping period and there being a potential for a misconnected block.

RaphiMC: Yeah right I forgot Via already only checks against blocks which could have connections. It shouldn't be that bad then

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/blockconnections/ConnectionData.java
```diff
@@ -73,8 +75,10 @@ public static void update(UserConnection user, Position position) throws Excepti
             }
 
             int newBlockState = handler.connect(user, pos, blockState);
-            if (newBlockState == blockState && blockConnectionProvider.storesBlocks()) {
-                continue;
+            if (newBlockState == blockState) {
+                if (inSync == null) inSync = blockConnectionProvider.storesBlocks(user, position);
```
kennytv, line 79: Code style here and in the other cases: expand the ifs with brackets and a new line

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/storage/BlockConnectionStorage.java
```diff
@@ -88,10 +94,23 @@ public void remove(int x, int y, int z) {
...
+        for (Position p : modified) {
+            if (Math.abs(pos.x() - p.x()) + Math.abs(pos.y() - p.y() + Math.abs(pos.z() - p.z())) < 2)
```
kennytv, line 103: Bad parentheses (?)
```suggestion
            if (Math.abs(pos.x() - p.x()) + Math.abs(pos.y() - p.y()) + Math.abs(pos.z() - p.z()) < 2)
```

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/packets/WorldPackets.java
kennytv: Since all you need here is the position, you can remove the wrapper calls below this and change it to
`Position pos = wrapper.passthrough(Type.POSITION)`

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/storage/BlockConnectionStorage.java
Pablete1234: Nice catch

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/storage/BlockConnectionStorage.java
Pablete1234: I'll be keeping it at 5, and increasing the distance to <= 2 (vs < 2). Even if you have ~250ms and are placing blocks at ~20 blocks per second, this has you covered, for anything past that, we'll just hope you aren't doing it near connected blocks and being cancelled by the server: it's just too specific of a case to warrant increasing this 5, given the potential performance implication.

Increasing the distance also solves another bug (placing a door, if the top-half connects something) because you place against the floor, but the top-half that needs to be removed is 2 blocks away from the clicked one, so pre-change it'd be broken.

Pablete1234:
Updated the pr based on the comments, and made a couple other tweaks of my own:
 - Block positions now use <= 2 taxicab distance. While this is generous, it can be useful if you place a door against the floor, as the recorded block would be the floor and the top-half (which could be removed if event is cancelled) would be 2 blocks away. Additionally it is more lenient if you ever were so laggy that 5 positions didn't fully cover what you've touched.
 - Since starting to dig is what adds a position, i've made it so the same position won't be saved twice. Reason i need to do it on digging-start is that's the only packet creative will send, and i'd rather not have to also listen to gamemode packets to know if i should listen to start digging or end digging.

RaphiMC (approved):
Looks good to me now

kennytv:
Can you elaborate the reproduction steps? I don't actually get the issue on either a 1.8 or 1.12 server with a 1.20 client using the default config and a simple block place listener that cancels the event

Pablete1234:
I'm reproducing this using 1.8 paper and a 1.18.2 client, make sure that the viaversion config for block storage is packet, not world

A simple plugin that denies block place events should do

Issue #2923, opened 2022-06-08
1.19 players are kicked when a plugin uses player.sendTitle()


tyler58546:
### `/viaversion dump` Output

https://dump.viaversion.com/d2dfceb1e4a63fa5879b9f725f2cd7cb5508280bd88f8e00296a69140c11e848

### Console Error

![Internal Exception: io.netty.handler.codec.DecoderException: Received unexpected null component](https://user-images.githubusercontent.com/16512508/172505365-e6dc0f13-e411-473b-aa66-535a41043cb5.png)
`Internal Exception: io.netty.handler.codec.DecoderException: Received unexpected null component`

### Bug Description

1.19 players are kicked when a plugin uses `player.sendTitle()`. Players on other versions are not kicked and receive the title as expected. Sending titles with the vanilla `/title` command works fine for 1.19 players.

### Steps to Reproduce

Use a plugin that runs `player.sendTitle()`

### Expected Behavior

The player should not be kicked

### Checklist

- [X] Via plugins are only running on **EITHER** the backend servers (e.g. Paper) **OR** the proxy (e.g. BungeeCord), **not on both**.
- [X] I have included a ViaVersion dump.
- [X] If applicable, I have included a paste (**not a screenshot**) of the error.
- [X] I have tried the latest build(s) from https://ci.viaversion.com/ and the issue still persists.

Barvalg:
![Platform: git--Paper--392%20%28MC%3A%201.15.2%29](https://img.shields.io/badge/Platform-git--Paper--392%20%28MC%3A%201.15.2%29-3498db?style=flat-square) </br>![ViaVersion (4.3.1): Even with master](https://img.shields.io/badge/ViaVersion%20(4.3.1)-Even%20withmaster-brightgreen?style=flat-square) </br>![ViaBackwards(4.3.0): Even with master](https://img.shields.io/badge/ViaBackwards%20(4.3.0)-Even%20with%20master-brightgreen?style=flat-square) </br>

Brokkonaut:
I have the same issue on a paper server when sending empty titles.

kennytv:
What's the expected behavior here? Remove the title or display an empty title?

TDMidkwhatisxd:
I can confirm it with betterRTP, after removing Title, it didnt kick the players anymore. Funny enough, bedrock Players dont experience THis issue

PR #3366, opened 2023-06-24
Handle more null TextComponent for 1.19 to 1.18.2

sandtechnology:
Through in the previous issue (#2923), the fix handle the component on title packet, but null component could also appear on action bar, teams and scoreboard objective packet and result in the same "Received unexpected null component" error, this PR handle those packets above and map it into empty component like the fix commit (https://github.com/ViaVersion/ViaVersion/commit/66b21431f22a62cc1f3d0a5cdd28bd9ea1e36944).


kennytv:
Remove the two passthroughAll calls, no need to read/handle anything past what you have to write in any individual handler

kennytv (approved):
Thanks

PR #3402, opened 2023-08-13
Optimize memory allocation for data palette

Pablete1234:
Improves memory allocation time when lots of players are being moved between worlds. The test setup is 80 players in 1.20, connected to a 1.8 server, being teleported at once from a world to another.

Current viaversion is using upwards of 20GB worth of memory allocations solely on DataPaletteImpl, with the applied optimizations it lowers to ~6GB~ 2.1GB (world with some empty regions) or even 800MB (world with mostly void).

Before: 
image
image

After:
image
image

The main  gains are:
 - byte[] instead of int[] for the chunk data, and nothing at all for empty chunks
   - Works as long as there's up to 256 different types of blocks
   - If there's over 256 block types, it'll migrate over to a short[] instead
   - Turns ~9.31GB of allocations into ~2.92GB, with some empty chunks 1.35GB, with mostly void 5MB
 - Initializing the palette to 16 instead of 8.
   - very few chunks use less than 8, so they almost always had to resize
   - Turns ~7.34GB of allocations into 1.58GB
   - Tried alternative parameters (32, 64, 128) but they didn't improve. Allocating more is a trade-off, some will enjoy it being available, but the baseline increases. 

initial | resizes (MB) | init (MB) | total (MB)
-- | -- | -- | --
8 | 7280 | 62 | 7342
16 | 1970 | 94 | 2064
32 | 1850 | 199 | 2049
64 | 1770 | 402 | 2172
128 | 2070 | 624 | 2694

- The inverse palette had behavior where bumping up the size caused an increase of initial (as expected) but the resizes were barely accounting for anything.

initial | resizes (MB) | init (MB) | total (MB)
-- | -- | -- | --
8 | 54 | 244 | 298
16 | 78 | 392 | 470
32 | 56 | 838 | 894

I'm unsure as to what causes this behavior, but i've left it in a way that the initial size is kept as it was (table size of 16, which is the same as it was with expected size of 8). The theory may be that most chunks have enough with 12 blocks.

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/minecraft/chunks/DataPaletteImpl.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/minecraft/chunks/DataPaletteImpl.java
@@ -41,7 +41,7 @@ public final class DataPaletteImpl implements DataPalette {
     }
 
     public DataPaletteImpl(final int valuesLength, final int initialSize) {
-        values = new ByteChunkData(valuesLength);
+        values = new EmptyChunkData(valuesLength);
         sizeBits = Integer.numberOfTrailingZeros(valuesLength) / 3;
         // Pre-size the palette array/map
         palette = new IntArrayList(initialSize);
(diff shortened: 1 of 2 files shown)


kennytv (approved):
Thanks, looks good; and thank you for the detailled summary!

RaphiMC:
This PR has some flaws in the assumptions it makes about the ID system. Negative IDs or integer IDs aren't handled at all causing any code which relies on them to break in unexpected ways (Both cases which the old code handled fine)

Pablete1234:
Hey @RaphiMC the only change in the system is to palette indexes. Block ids, stored in the palette, can continue to be negative and int-based, while the palette indexes in the chunk block data is now byte/short. The amount of different blocks in a chunk section is at most 16x16x16, each different block type needs a new palette index, which limits the possible max palette index at 2^12 while short is big enough for up to 2^32.
Am i missing something here?

PR #3407, opened 2023-08-19
Fix placing blocks inside the player being cancelled often

Pablete1234:
There's a piece of code to prevent blocks being able to place inside yourself, this however has a few flaws, and blocks are place-prevented when they shouldn't.

Current viaversion logic is to prevent any solid block placement, except for banners and signs. This has both false-positives and false-negatives, eg:
 - Pressure plates can't be placed (solid material) even tho they have no collision box
 - Ladders can always be placed (non-solid material) even if standing where it'll end up.

The new approach checks for actual collision between the players' bounding box and the block they just placed, which re-allows placing stuff like fences, slabs, glass panes, or doors in the block you're standing in, given that you're not in the place where they'll be placed.

Here's a gif with a bunch of block-places that are currently prevented with viaversion (1.8 server, 1.18 client):
block_placement

I've tested these changes with sportpaper 1.8 only, will follow-up with some tests in paper 1.8 and paper 1.12 to make sure the methods exist across all versions. If the methods aren't found, it'll fall-back to the old approach, and i've added pressure plates to list of allowed as an easy fix too.


kennytv (approved):
Thanks!

PR #3443, opened 2023-09-25
Fix disconnecting error for unknown mobs with metadata

Unoqwy:
Fixes a bug where entities whose metadata produces errors that should only be logged produce an IndexOutOfBoundsException.

The error comes from an attempted micro-optimization using an index that:
* goes to -1 on second metadata error (thus the out of bounds)
* is invalid as calls within the loop may add/remove elements to the list making index-based removal undefined behavior

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/rewriter/EntityRewriter.java
+++ b/common/src/main/java/com/viaversion/viaversion/rewriter/EntityRewriter.java
@@ -104,11 +104,10 @@ public abstract class EntityRewriter<C extends ClientboundPacketType, T extends 
     public void handleMetadata(final int entityId, final List<Metadata> metadataList, final UserConnection connection) {
         final TrackedEntity entity = tracker(connection).entity(entityId);
         final EntityType type = entity != null ? entity.entityType() : null;
-        int i = 0; // Count index for fast removal
         for (final Metadata metadata : metadataList.toArray(EMPTY_ARRAY)) { // Copy the list to allow mutation
             // Call handlers implementing the old handleMetadata
             if (!callOldMetaHandler(entityId, type, metadata, metadataList, connection)) {
-                metadataList.remove(i--);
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/viaversion/viaversion/rewriter/EntityRewriter.java
```diff
@@ -104,11 +104,10 @@ public void registerFilter(MetaFilter filter) {
     public void handleMetadata(final int entityId, final List<Metadata> metadataList, final UserConnection connection) {
         final TrackedEntity entity = tracker(connection).entity(entityId);
         final EntityType type = entity != null ? entity.entityType() : null;
-        int i = 0; // Count index for fast removal
         for (final Metadata metadata : metadataList.toArray(EMPTY_ARRAY)) { // Copy the list to allow mutation
             // Call handlers implementing the old handleMetadata
             if (!callOldMetaHandler(entityId, type, metadata, metadataList, connection)) {
-                metadataList.remove(i--);
+                metadataList.remove(metadata);
```
kennytv, line 110: The toArray call means only the original list that is not being iterated on will be modified. Technically two metadata objects can be equal and result in the wrong one being removed if state is involved, so the easier fix is to just remove the -- in this line

Unoqwy: The issue is `metadataList` is modified (with remove/add operations) within `callOldMetaHandler` call (removing by object already btw). So the index of iteration here may be different from what the list actually holds.

Unoqwy:
CI failing seems to be independent from my doing and related to a change made an hour ago (87db6fc). Lmk if you want help fixing that

Unoqwy:
Hey, is there anything blocking the PR besides the concern that was addressed in the review?

kennytv (approved):

PR #3490, opened 2023-10-21
Implemented support for chat signing

RaphiMC:
Allows newer clients to chat on older servers which enforce signed chat. Also allows joining onto servers which require a secure game profile. Needs the client's private signing key. To use this one needs to put the ChatSession_X classes into a UserConnection before the login process.


FlorianMichael (approved):

RaphiMC:
Done with the changes

kennytv (approved):

PR #3501, opened 2023-10-24
Add own permission for every sub command

FlorianMichael:
Implemented changes from https://github.com/ViaVersion/ViaVersion/issues/2872.

The root/global permission remains as **viaversion.admin**, while the sub-command permissions are structured as **viaversion.admin.<commandname>**.

I removed the help sub-command, as the root command will now exclusively display the commands accessible to a user. Additionally, I made enhancements to Bukkit's plugin.yml to ensure that all sub-commands are correctly displayed when using **/help viaversion**.


common/src/main/java/com/viaversion/viaversion/commands/ViaCommandHandler.java
```diff
@@ -69,6 +69,17 @@ public ViaSubCommand getSubCommand(String name) {
 
     @Override
     public boolean onCommand(ViaCommandSender sender, String[] args) {
+        final AtomicBoolean hasPermissions = new AtomicBoolean(sender.hasPermission("viaversion.admin"));
+        commandMap.values().forEach(command -> {
```
kennytv, line 73: Use a normal for loop and boolean with a break

kennytv (approved):

PR #3502, opened 2023-10-24
Remove permission requirements for root command

FlorianMichael:
Slightly fix to my last PR ™️


kennytv (changes requested):
This would break tab completion permission checks

kennytv (approved):

PR #3504, opened 2023-10-25
Added Runtime-Information to dump

FlorianMichael:
See https://github.com/ViaVersion/ViaVersion/pull/2451


kennytv:
On second thought, that information isn't actually useful usually

PR #3587, opened 2023-12-16
Fix skin layers breaking for >= 1.20.2 clients on < 1.20.2 velocity proxies

Aeltumn:
This fixes an issue MCC Island (and likely other servers) are having where skin layers go missing when players switch servers. We run ViaVersion only on the proxy. Without this fix skin layers break every server switch, seems completely fixed after.

I don't know enough about Via's workings to fully understand why this fixes it but my guess is that there's some issue with the packet arriving in the wrong phase when the client doesn't go into the configuration phase first? Maybe this needs a better fix.

Found this fix because were running a very specific developer build that had 1.20.2 support but without this issue, and I reverted the one commit that made any changes and this was the only part of that commit that has any effects.


kennytv (approved):

radiant-ai:
I just wonder if this could be related or be same issue as this one:
https://github.com/PaperMC/Waterfall/issues/835

PR #3662, opened 2024-01-21
Fix missing handling of text components in 1.11.1>1.12 chat item rewriter

LeonTG:
I never noticed that this was not handled by ViaVersion until upgrading to 1.20.3 and above

1.20.3 changed it so a client will be kicked if an invalid chat json text is sent to the client. 

ViaVersion did not check for hover events using a JsonObject with a text element inside, and then made the hover event an empty array instead, which on 1.20.3 and above, would kick the client for the reason seen here: image

I have tested this and it seems to have fixed the issue, now sending a item or entity hoverEvent on a 1.20.3+ client on a 1.8-1.11 server no longer kicks me

Hopefully I described the issue well and submitted an adequate fix for it! Let me know if anything is unclear or if there is changes you'd want me to do to the fix!

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_12to1_11_1/ChatItemRewriter.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_12to1_11_1/ChatItemRewriter.java
@@ -42,6 +42,15 @@ public class ChatItemRewriter {
                             if (value.isJsonPrimitive() && value.getAsJsonPrimitive().isString()) {
                                 String newValue = indexRemoval.matcher(value.getAsString()).replaceAll("");
                                 hoverEvent.addProperty("value", newValue);
+                            } else if (value.isJsonObject() && value.getAsJsonObject().has("text")) {
+                                JsonObject newObject = (JsonObject) value;
+                                JsonElement textElement = newObject.get("text");
+
+                                if (textElement.isJsonPrimitive() && textElement.getAsJsonPrimitive().isString()) {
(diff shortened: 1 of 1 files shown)


FlorianMichael (approved):
Didn't had the time to look at it in the last weeks, sorry about the delay! 

While this seems to be valid, we should take a closer look at all ChatItemRewriter when properly migrating to MCStructs / adding all missing conversions.

Thanks!

FlorianMichael:
Also could you please provide the nbt / command you used to fix/debug this? I might found some issues in the existing code as well

FlorianMichael:
Are you able to reproduce this kick using /tellraw and your json? I can't reproduce this using latest ViaProxy, 1.20.4 client and 1.8 server, same for ViaVersion being installed on the spigot server.

LeonTG:
> Are you able to reproduce this kick using /tellraw and your json? I can't reproduce this using latest ViaProxy, 1.20.4 client and 1.8 server, same for ViaVersion being installed on the spigot server.

I believe /tellraw would change the formatting of the hover event back to a basic string primitive, which ViaVerison already handles

The way this json gets sent in my case is using the Spigot components API

GWServer:
The update you published today did not fix this. Check out my thread. I described it in detail.
https://github.com/ViaVersion/ViaVersion/issues/3726

FlorianMichael:
Thanks for the PR, I decided to use MCStructs for the conversion to fix even more issues the current code has (For example you can split the nbt using translate and extra or you can insert multiple arrays into each other), see my PR: https://github.com/ViaVersion/ViaVersion/pull/3740

PR #3691, opened 2024-02-13
Added ProtocolVersionRange API

FlorianMichael:
This PR adds a new ProtocolVersionRange API which can be used to define and compare a range/set of ProtocolVersions, I also renamed the existing VersionRange class file

### DIFF
--- a/api-legacy/src/main/java/us/myles/ViaVersion/api/protocol/ProtocolVersion.java
+++ b/api-legacy/src/main/java/us/myles/ViaVersion/api/protocol/ProtocolVersion.java
@@ -23,7 +23,7 @@
 package us.myles.ViaVersion.api.protocol;
 
 import com.google.common.base.Preconditions;
-import com.viaversion.viaversion.api.protocol.version.VersionRange;
+import com.viaversion.viaversion.api.protocol.version.SubVersionRange;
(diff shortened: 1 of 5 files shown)


api-legacy/src/main/java/us/myles/ViaVersion/api/protocol/ProtocolVersion.java
```diff
@@ -42,23 +42,23 @@ public class ProtocolVersion {
     private static final Int2ObjectMap<ProtocolVersion> versions = new Int2ObjectOpenHashMap<>();
     private static final List<ProtocolVersion> versionList = new ArrayList<>();
 
-    public static final ProtocolVersion v1_4_6 = register(51, "1.4.6/7", new VersionRange("1.4", 6, 7));
+    public static final ProtocolVersion v1_4_6 = register(51, "1.4.6/7", new SubVersionRange("1.4", 6, 7));
```
Kichura, line 45: Wasn't 1.4.6 - 1.6.4 cut in [this commit](https://github.com/ViaVersion/ViaVersion/commit/22bd350e35880b345203ce4e93078a5ee43e5e24#diff-c50e2b3c6c8a671c432d4b12aab95cf663bb4a3ab74eb73eec6d06b81f26526d) ?

Kichura (commented):
T

Issue #686, opened 2017-06-12
Crashing Pipeline and NPE on Protocol selection causing nobody to be able to join


geNAZt:
### What is the output link of /viaversion dump? 
-

### Are you using any additional software like BungeeCord? If so, what software and version? (Please list your plugins as well)
Yes, latest

### How does this error happen? login? Using an item? 
Starting the server and first join

### Is there an error in the console? Use pastebin.com. Is there a kick message?
https://pastebin.com/raw/pys4P2gG

MylesIsCool:
I've changed it so ViaVersion is completely inactive at boot time, this will stop these errors and mean that 1.8 can join if you're using a 1.8 server (but other versions cannot, though there is no change here).

I'm not sure if you wanted support for other versions joining during boot, this isn't possible with how the server declares what protocol version it is. (As it's done after boot phase)

PR #3702, opened 2024-02-16
Fix typo in ViaVersion's description

Kichura:
Synced from [ViaBackwards' #686](https://github.com/ViaVersion/ViaBackwards/pull/686), ViaRewind doesn't require this treatment though.


FlorianMichael:
It's not a typo, I think we should change it in ViaRewind to also use "Allow"

Kichura:
fair then, closing both.

PR #3710, opened 2024-02-24
Fix wrong join game packet handling

LemonCaramel:
Fixes the issue where the Portal cooldown always becomes 0.


kennytv (approved):
Thanks!

PR #3712, opened 2024-02-25
Fix particle doesn't use new mapping id (1.20.5)

LemonCaramel:
The particle id was not converted.


common/src/main/java/com/viaversion/viaversion/protocols/protocol1_20_5to1_20_3/rewriter/BlockItemPacketRewriter1_20_5.java
```diff
@@ -91,6 +91,7 @@ public void registerPackets() {
                 particle.add(Type.ITEM1_20_2, item);
             }
 
+            particle.setId(mappings.getNewId(particleId));
```
FlorianMichael, line 94: This can be done when creating the Particle instance in line 84

FlorianMichael (approved):
Thanks

kennytv (approved):

PR #3728, opened 2024-02-28
Tidy up fabric.mod.json file of fabric platform

FlorianMichael:
ViaVersion will only work with ViaFabric so we should require it.

### DIFF
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -16,13 +16,11 @@
     "_MylesC",
     "creeper123123321",
     "Gerrygames",
-    "KennyTV",
+    "kennytv",
     "Matsv"
   ],
   "depends": {
-    "fabricloader": ">=0.14.0"
-  },
-  "recommends": {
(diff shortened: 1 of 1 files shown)


FlorianMichael:
On ViaFabricPlus ViaVersion will only do something if putted into the jars folder, it won't do anything as a Mod itself.

HowardZHY:
> On ViaFabricPlus ViaVersion will only do something if putted into the jars folder, it won't do anything as a Mod itself.

Won't that cause something when jar in jar?

FlorianMichael:
No, VFP allows you to replace the jar in jar packaged VV jar on load using the jars folder, but VFP will ignore any Via* jar in the mods folder

HowardZHY:
> > On ViaFabricPlus ViaVersion will only do something if putted into the jars folder, it won't do anything as a Mod itself.
> 
> Won't this cause something wrong when jar in jar?

Fine, it's viaversion-common and it's fabric.mod.json isn't this

Kichura (approved):

PR #3729, opened 2024-02-28
Add classic version range to VersionType#CLASSIC javadoc

FlorianMichael:


FlorianMichael:
> Classic 0.0.2a does not have multiplayer

Alpha 1.0.0 didn't had too but we included it.

PR #3738, opened 2024-03-06
Implement more 24w09a item components

DrexHD:
Adds support for `MAP_COLOR`, `MAP_ID`, `MAP_DECORATIONS` and `PROFILE`


kennytv:
Thanks for the PR - I had to iron some things out, so I merged manually in https://github.com/ViaVersion/ViaVersion/commit/386495e1681cff47e82712d300b1449f39248ffc. Map data for example is still written as NBT, otherwise just smaller things. Fyi, the rest of this update will most likely take place in the semi-private repo, but I decided to push this one public still

PR #3741, opened 2024-03-09
Replaced printStackTrace usages with proper logging

FlorianMichael:
Added proper logging with error messages in most places so we can identify errors easier.


kennytv:
Should always be severe, not just a warning

kennytv (approved):

PR #3754, opened 2024-03-23
Allow platforms to use the mapping system

FlorianMichael:
Allows platforms to create their own MappingDataLoader and to override the loading behaviour in MappingDataBase.

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/data/MappingDataLoader.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/data/MappingDataLoader.java
@@ -51,17 +51,20 @@ import java.util.Map;
 import java.util.stream.Collectors;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
-public final class MappingDataLoader {
+public class MappingDataLoader {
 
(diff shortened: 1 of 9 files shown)


api/src/main/java/com/viaversion/viaversion/api/data/MappingDataLoader.java
```diff
@@ -51,17 +51,20 @@
...
     private static final byte SHIFTS_ID = 1;
     private static final byte CHANGES_ID = 2;
     private static final byte IDENTITY_ID = 3;
-    private static boolean cacheValid = true;
 
-    public static void clearCache() {
+    public static final MappingDataLoader INSTANCE = new MappingDataLoader();
+
+    private final Map<String, CompoundTag> MAPPINGS_CACHE = new HashMap<>();
+    private final TagReader<CompoundTag> MAPPINGS_READER = NBTIO.reader(CompoundTag.class).named();
```
kennytv, line 64: Mappings cache needs to be renamed, and mappings reader can stay static

FlorianMichael: Fixed

kennytv (approved):

PR #3763, opened 2024-03-29
Move chat registry dumps into nbt files

FlorianMichael:


FlorianMichael:
Reasons for this:
- ViaVersion always moves out data dumps (especially nbt) into own files
- Allows users to edit them if needed for some reason
- Cleaner than having random SNBT at the top of the class

kennytv (approved):

PR #3765, opened 2024-03-30
Rewrite 1.12.2->1.13 ComponentRewriter

FlorianMichael:
Move away from ViaNBT's SNBT parsing and replaces the code with proper MCStructs parsing

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/data/ComponentRewriter1_13.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/data/ComponentRewriter1_13.java
@@ -17,14 +17,12 @@
  */
 package com.viaversion.viaversion.protocols.protocol1_13to1_12_2.data;
 
-import com.github.steveice10.opennbt.stringified.SNBT;
 import com.github.steveice10.opennbt.tag.builtin.CompoundTag;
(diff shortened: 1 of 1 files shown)


FlorianMichael:
Basically https://github.com/ViaVersion/ViaVersion/pull/3740 but for the next protocol, at some point I'll maybe need to implement proper chat item rewriter into all protocols so chat items are working.

kennytv:
Any examples of what this actually fixes?

FlorianMichael:
The findItemNBT would cause the same issue as in https://github.com/ViaVersion/ViaVersion/pull/3740 since it doesn't handle most cases. I also replaced ViaNBT's SNBT parser since it doesn't handle version differences which exists properly, I have no compound at the moment but all changes between 1.12.2 and 1.13 wouldn't be probably handled without MCStructs. We are already using MCStructs when parsing SNBT/Text components everywhere else so there is no reason for us to not do it in here too, I also tried some cursed nbts and they are working with the PR's changes.

FlorianMichael:
Here is an example nbt which will cause the old code to crash:
`/tellraw @a {"text":"Guter Text","hoverEvent":{"action":"show_item","value":{"text": "{", "extra": [{"text": "id:stone,Damage:1,Count:1}"}]}}}`

kennytv (approved):

PR #3774, opened 2024-04-05
Remove MetaType1_8#NonExistent

FlorianMichael:
We don't need this and can just check if **getOldType()** is null


kennytv:
Should go into https://github.com/ViaVersion/ViaVersion/pull/3775

kennytv (approved):

PR #3778, opened 2024-04-06
Avoid `NullPointerException` in case Via.getManager().addEnableListener() is called after the manager is initialized

wtlgo:
If an external plugin calls `Via.getManager().addEnableListener()` after `Via.getManager().init()` has been called, it causes a `NullPointerException`. In this pull request, to prevent this issue, if the manager is already initialized, it immediately executes the listener instead of raising an exception and causing the external plugin to fail during its loading process.


wtlgo:
Sorry, forked a wrong branch

PR #3779, opened 2024-04-06
Preventing `NullPointerException` when calling `Via.getManager().addEnableListener()` after manager is initialized

wtlgo:
If an external plugin calls `Via.getManager().addEnableListener()` after `Via.getManager().init()` has been called, it causes a `NullPointerException`. In this pull request, to prevent this issue, if the manager is already initialized, it immediately executes the listener instead of raising an exception and causing the external plugin to fail during its loading process.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/ViaManagerImpl.java
+++ b/common/src/main/java/com/viaversion/viaversion/ViaManagerImpl.java
@@ -329,7 +329,11 @@ public class ViaManagerImpl implements ViaManager {
      * @param runnable runnable to be executed
      */
     public void addEnableListener(Runnable runnable) {
-        enableListeners.add(runnable);
+        if (enableListeners != null) {
+            enableListeners.add(runnable);
+        } else {
+            runnable.run();
+        }
     }
 
     @Override


kennytv:
What's the use case here? If any of our plugins call this too late, this *should* mean failure (though ofc the exception could be better), also just being internal api

wtlgo:
> What's the use case here?

E.g. If the external plugin is loaded by another plugin during the loading phase via let's say `Bukkit.getPluginManager().loadPlugin()` it is quite literally impossible to manage the external plugin to call `addEnableListener` before ViaVersion calls `init` even though it is not misconfiguration, as all it tries to do is not to run its ViaVersion-dependent init process before ViaVersion is ready, and as result both ViaVersion and plugin in question would function just fine.

Also, this is *not* "just internal API", as it is [exposed by `com.viaversion:viaversion-api`](https://jd.viaversion.com/com/viaversion/viaversion/api/ViaManager.html#addEnableListener-java.lang.Runnable-).

kennytv:
Is that external plugin one of ours, or one you can change the method call in? The current name doesn't really imply any immediate execution and even in your scenario should still provide a warning. Assuming this is about ViaBackwards or Rewind, this will lead to peformance issues, possibly even more so in the future, due to missing data availability (vs. keeping order via something like the `add-plugin` startup flag)

FlorianMichael:
Maybe we could consider moving this into common since it's not really API other plugins should use?

Kichura:
Closing this because OP of this PR has made no responses for a while now, if you have any updates to give then feel free to make a reply or make new PR that superseds this.

PR #3780, opened 2024-04-09
Fix 1.13 -> 1.12.2 block state metadata translation

RaphiMC:
Fixes 1.13 -> 1.12.2 block state metadata translation by applying the transforms in the correct order (Previously it would first try to get the 1.13 block state with the raw 1.12.2 value instead of converting the 1.12.2 value to a proper 1.12.2 blockstate first)


RaphiMC:
Wrong branch

PR #3781, opened 2024-04-09
Fix 1.13 -> 1.12.2 block state metadata translation

RaphiMC:
Fixes 1.13 -> 1.12.2 block state metadata translation by applying the transforms in the correct order (Previously it would first try to get the 1.13 block state with the raw 1.12.2 value instead of converting the 1.12.2 value to a proper 1.12.2 blockstate first)


FlorianMichael (approved):

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/metadata/MetadataRewriter1_13To1_12_2.java
```diff
@@ -53,28 +59,13 @@ protected void registerRewrites() {
...
-        });
-
         filter().type(EntityTypes1_13.EntityType.WOLF).index(17).handler((event, meta) -> {
             // Handle new colors
             meta.setValue(15 - (int) meta.getValue());
         });
 
         filter().type(EntityTypes1_13.EntityType.ZOMBIE).addIndex(15); // Shaking
 
-        filter().type(EntityTypes1_13.EntityType.MINECART_ABSTRACT).index(9).handler((event, meta) -> {
```
FlorianMichael, line 71: Needs to be added back since it's not an block state type.

RaphiMC: Fixed

kennytv (approved):

PR #3782, opened 2024-04-09
Handle block states optional in <1.19.4

FlorianMichael:
This PR fixes block state reading in versions below 1.19.4 where there was only a block state type which is optional, we can probably also rename some of the meta type fields with ViaVersion 5.

1.19.3 code:
image

1.19.4 code:
image

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/metadata/MetadataRewriter1_13To1_12_2.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/metadata/MetadataRewriter1_13To1_12_2.java
@@ -40,7 +40,11 @@ public class MetadataRewriter1_13To1_12_2 extends EntityRewriter<ClientboundPack
     protected void registerRewrites() {
         filter().mapMetaType(typeId -> Types1_13.META_TYPES.byId(typeId > 4 ? typeId + 1 : typeId));
         filter().metaType(Types1_13.META_TYPES.itemType).handler(((event, meta) -> protocol.getItemRewriter().handleItemToClient(meta.value())));
-        filter().metaType(Types1_13.META_TYPES.blockStateType).handler(((event, meta) -> meta.setValue(WorldPackets.toNewId(meta.value()))));
+        filter().metaType(Types1_13.META_TYPES.blockStateType).handler(((event, meta) -> {
+            if ((int) meta.value() != 0) {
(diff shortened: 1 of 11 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/protocol1_18to1_17_1/packets/EntityPackets.java
```diff
@@ -94,7 +94,7 @@ protected void registerRewrites() {
             }
         });
 
-        registerMetaTypeHandler(Types1_18.META_TYPES.itemType, null, null, null);
+        registerMetaTypeHandler(Types1_18.META_TYPES.itemType,null, null);
```
kennytv, line 97: Missing space

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/metadata/MetadataRewriter1_13To1_12_2.java
```diff
@@ -40,7 +40,11 @@ public MetadataRewriter1_13To1_12_2(Protocol1_13To1_12_2 protocol) {
     protected void registerRewrites() {
         filter().mapMetaType(typeId -> Types1_13.META_TYPES.byId(typeId > 4 ? typeId + 1 : typeId));
         filter().metaType(Types1_13.META_TYPES.itemType).handler(((event, meta) -> protocol.getItemRewriter().handleItemToClient(meta.value())));
-        filter().metaType(Types1_13.META_TYPES.blockStateType).handler(((event, meta) -> meta.setValue(WorldPackets.toNewId(meta.value()))));
+        filter().metaType(Types1_13.META_TYPES.blockStateType).handler(((event, meta) -> {
+            if ((int) meta.value() != 0) {
```
kennytv, line 44: put meta.value into a variable

kennytv (approved):
Accepted because it's cleaner, but this doesn't actually "fix" anything, 0 is always 0

FlorianMichael:
"Fix" is the wrong word, but why would we handle and/or name types wrong, it's just cleaner and prevents confusion when looking at old code.

common/src/main/java/com/viaversion/viaversion/protocols/protocol1_13to1_12_2/metadata/MetadataRewriter1_13To1_12_2.java
FlorianMichael: Fixed

PR #3786, opened 2024-04-15
Add IdAndData util and replace conversions

FlorianMichael:
Basically ViaBackwards's Block class but with some additions and error handlers for ViaLegacy/ViaRewind. This replaces https://github.com/ViaVersion/ViaLegacy/blob/main/src/main/java/net/raphimc/vialegacy/api/model/IdAndData.java and https://github.com/ViaVersion/ViaBackwards/blob/dev/common/src/main/java/com/viaversion/viabackwards/utils/Block.java and also some otherwise dirty conversions in existing protocols.

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/util/IdAndData.java
+++ b/api/src/main/java/com/viaversion/viaversion/util/IdAndData.java
@@ -31,6 +31,10 @@ public class IdAndData {
     private int id;
     private byte data;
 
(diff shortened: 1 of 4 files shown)


kennytv:
Compressed -> Encoded (or raw)

FlorianMichael:
Changed.

kennytv (approved):

PR #3787, opened 2024-04-16
Replace incoming/outgoing usages with client/server bound

FlorianMichael:
Since VV supports both server and client side, it makes more sense to use client/server bound as wording since incoming/outgoing will always be wrong on one side. Also changed 1.12->1.13 handlers to print the correct channel name.


kennytv:
No space between the two parts

FlorianMichael:
Fixed.

kennytv (approved):

PR #3793, opened 2024-04-19
Fix 1.16->1.15.2 item attribute uuid

AJ-Ferguson:
Fixes an item desync between client and server


common/src/main/java/com/viaversion/viaversion/protocols/protocol1_16to1_15_2/packets/InventoryPackets.java
```diff
@@ -203,7 +203,7 @@ public static void oldToNewAttributes(Item item) {
             NumberTag leastTag = attribute.getNumberTag("UUIDLeast");
```
FlorianMichael, line 203: I suppose if we want to remove data from the old protocol, you should use remove() for both UUIDLeast and UUIDMost?

AJ-Ferguson: Makes sense to me. Added

FlorianMichael (approved):

kennytv (approved):
Thank you!

PR #3812, opened 2024-04-28
Expose ComponentRewriter1_20_5#toTag

FlorianMichael:
Later other projects might need to access the conversion done in the ComponentRewriter1_20_5


kennytv:
Already made it public

PR #3819, opened 2024-05-01
Add optional registry utils to TagUtil

FlorianMichael:
Required in ViaBackwards where rewriting of chat types requires the handling to be optional since chat_types are optional and can be skipped by the server


kennytv:
Should add a nullable default param instead with the original method as an overload to null

kennytv (approved):

PR #3827, opened 2024-05-05
Add alpha to particle colors in 1.20.3->.5

FlorianMichael:
Fixes entity effects not showing because the alpha channel was missing (which got added in 1.20.5 and was previously always 1, so we are using 1.20.4's default value here). Area effect clouds are still broken, but that's more likely an issue with the handler itself.


kennytv (approved):

FlorianMichael:
Yeah as I said there are still things wrong, but the change in my pull request is correct and was missing.

FlorianMichael:
Yes it is, you basically just repeated your previous message.

PR #3834, opened 2024-05-08
Handle invalid disconnect message in 1.8->1.9

FlorianMichael:
This is a dirty fix for a change GSON made to its internal component parsing (see https://github.com/Lenni0451/MCStructs/issues/4). I'm not proud of this change but it's better than the old component conversion (where there wouldn't be a disconnect message at all) so until MCStructs fixes this, it's probably our best way to go for now...

Closes https://github.com/ViaVersion/ViaFabricPlus/issues/408 (can be tested by joining the mentioned server before and after the PR)


kennytv (approved):
oh well

PR #3843, opened 2024-05-11
Rework logging inside protocols, add Protocol#getLogger()

FlorianMichael:
This PR is reworking some of the current logging done in protocols, we now have protocol specific loggers which can automatically append the protocol direction to every message which makes it easier to make log messages without having the need to include the version into the message, in addition to that, AbstractViaConfig#isSuppressConversionWarnings now automatically checks for debug mode being enabled.


common/src/main/java/com/viaversion/viaversion/util/LogUtil.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    public void conversionWarning(Class<? extends Protocol> protocol, Supplier<String> message, Throwable t) {
+        if (!Via.getConfig().isSuppressConversionWarnings() || Via.getManager().isDebug()) {
+            logger.log(Level.WARNING, toNiceName(protocol) + ": " + message.get(), t);
+        }
+    }
+
+    // ---------------------------------------------------------------------------------
```
kennytv, line 58: No need for a separator, the class is simple enough

kennytv (commented):
Having a createConversionLogger method in AbstractProtocol that is put in a field there, `ViaLogger extends Logger` where you override log(LogRecord) might be better

common/src/main/java/com/viaversion/viaversion/util/ProtocolLogger.java
```diff
@@ -0,0 +1,61 @@
...
+    public void warning(final String msg) {
+        logger.warning(formatMessage(msg));
+    }
+
+    public void severe(final String msg) {
+        logger.severe(formatMessage(msg));
+    }
+
+    private String formatMessage(final String msg) {
+        return "(" + ProtocolUtil.toNiceName(protocol) + ") " + msg;
```
kennytv, line 59: The prefix should be stored as a field and the protocol field removed

common/src/main/java/com/viaversion/viaversion/protocols/v1_10to1_11/Protocol1_10To1_11.java
```diff
@@ -41,8 +39,11 @@
 import com.viaversion.viaversion.protocols.v1_9_1to1_9_3.packet.ServerboundPackets1_9_3;
 import com.viaversion.viaversion.rewriter.SoundRewriter;
 import com.viaversion.viaversion.util.Pair;
+import com.viaversion.viaversion.util.ProtocolLogger;
 
 public class Protocol1_10To1_11 extends AbstractProtocol<ClientboundPackets1_9_3, ClientboundPackets1_9_3, ServerboundPackets1_9_3, ServerboundPackets1_9_3> {
+
+    public static final ProtocolLogger LOGGER = new ProtocolLogger(Protocol1_10To1_11.class);
```
kennytv, line 46: These should be moved to instance fields in AbstractProtocol with a getter there

common/src/main/java/com/viaversion/viaversion/protocols/v1_12_2to1_13/data/ParticleRewriter.java
```diff
@@ -96,7 +95,7 @@ public class ParticleRewriter {
 
     public static Particle rewriteParticle(int particleId, Integer[] data) {
         if (particleId >= particles.size()) {
-            Via.getPlatform().getLogger().severe("Failed to transform particles with id " + particleId + " and data " + Arrays.toString(data));
+            Protocol1_12_2To1_13.LOGGER.severe("Failed to transform particles with id " + particleId + " and data " + Arrays.toString(data));
```
kennytv, line 98: You can remove this error and just ignore it

kennytv (changes requested):

common/src/main/java/com/viaversion/viaversion/util/ProtocolLogger.java
FlorianMichael: Good point, thank you.

common/src/main/java/com/viaversion/viaversion/protocols/v1_12_2to1_13/data/ParticleRewriter.java
FlorianMichael: Is there any specific reason for this except not having a static LOGGER instance? (Which we now need anyway in this protocol)

common/src/main/java/com/viaversion/viaversion/protocols/v1_10to1_11/Protocol1_10To1_11.java
FlorianMichael: Done, I'll work on using Protocol#getLogger() in rewriter bases later

FlorianMichael:
All usages of the global logger inside of protocol related code should be replaced now, I wasn't sure for AbstractProtocol#printRemapError (since that would change the message). Some protocols need static context but there's not much we can do about, I'll work later on ViaBackwards pull requests to update API usage if you approve.

common/src/main/java/com/viaversion/viaversion/protocols/v1_12_2to1_13/data/ComponentRewriter1_13.java
```diff
@@ -54,8 +55,8 @@ protected void handleHoverEvent(UserConnection connection, JsonObject hoverEvent
         try {
             tag = ComponentUtil.deserializeLegacyShowItem(value, SerializerVersion.V1_12);
         } catch (Exception e) {
-            if (!Via.getConfig().isSuppressConversionWarnings() || Via.getManager().isDebug()) {
-                Via.getPlatform().getLogger().log(Level.WARNING, "Error reading 1.12.2 NBT in show_item: " + value, e);
+            if (!Via.getConfig().isSuppressConversionWarnings()) {
+                Protocol1_12_2To1_13.LOGGER.log(Level.WARNING, "Error reading NBT in show_item: " + value, e);
```
kennytv, line 59: This has a protocol instance it can use

kennytv (approved):

PR #3849, opened 2024-05-13
Fixup legacy entity type hierarchy and namings

FlorianMichael:
This PR fixes and updates entity types for pre 1.14 protocols, following changes have been made:
- Change naming and order of entities to be consistent with modern (1.14+) entity type enums
- Fixed all hierarchy issues (1.12- diffs have been created using a dumping tool I wrote)
- Cleanup code in general
- Add entity/(object) types for 1.8 and 1.9 as well
- Fixup MetaIndex in 1.8 protocol to use correct entity types

Would appreciate a review (especially on my changes in the 1.8->1.9 protocol itself). I also noticed that we don't do entity validation in protocols at all (the server could send 1.9 entity types and the 1.8->1.9 protocol would accept this, same for newer versions like 1.20.5). This is something I would like to add, also some enums aren't used now but I would keep them anyway for consistency.


api/src/main/java/com/viaversion/viaversion/api/minecraft/entities/EntityTypes1_13.java
```diff
@@ -29,201 +29,194 @@
...
+        EXPERIENCE_BOTTLE(76, PROJECTILE),
+        POTION(77, PROJECTILE),
+        FISHING_BOBBER(93, ENTITY),
+
+        ABSTRACT_ARROW(ENTITY),
+        ARROW(2, ABSTRACT_ARROW),
+        SPECTRAL_ARROW(68, ABSTRACT_ARROW),
+        TRIDENT(94, ABSTRACT_ARROW),
+
+        ABSTRACT_FIREBALL(ENTITY),
```
kennytv, line 84: Note that this is a bad mcp name, these aren't fireballs but Mojang `AbstractHurtingProjectile`, or later fixed in mcp as `DamagingProjectileEntity`

api/src/main/java/com/viaversion/viaversion/api/minecraft/entities/EntityTypes1_10.java
```diff
@@ -234,13 +264,12 @@ public EntityType getType() {
...
-        public static Optional<ObjectType> findById(int id) {
-            if (id == -1)
-                return Optional.empty();
+        public static Optional<ObjectType> findById(final int id) {
+            if (id == -1) return Optional.empty();
             return Optional.ofNullable(TYPES.get(id));
         }
 
-        public static Optional<EntityType> getPCEntity(int id) {
+        public static Optional<EntityType> getEntityType(final int id) {
```
kennytv, line 272: The optional returning methods could just be made nullable instead

api/src/main/java/com/viaversion/viaversion/api/minecraft/entities/EntityTypes1_13.java
FlorianMichael: They have the same name in modern enums, should I change all namings then?

kennytv: There's no hurting projectile type in those, that's the *actual* abstract fireball, which was added in 1.14

api/src/main/java/com/viaversion/viaversion/api/minecraft/entities/EntityTypes1_10.java
FlorianMichael: Yup thanks, fixed it.

api/src/main/java/com/viaversion/viaversion/api/minecraft/entities/EntityTypes1_13.java
FlorianMichael: Ah I see, changed the name to HURTING_PROJECTILE

common/src/main/java/com/viaversion/viaversion/protocols/v1_11_1to1_12/rewriter/EntityPacketRewriter1_12.java
```diff
@@ -81,7 +81,7 @@ protected void registerRewrites() {
             }
         });
 
-        filter().type(EntityTypes1_12.EntityType.EVOCATION_ILLAGER).index(12).toIndex(13); // Aggressive
+        filter().type(EntityTypes1_12.EntityType.EVOKER).index(12).toIndex(13); // Aggressive
```
kennytv, line 84: This should be ABSTRACT_ILLAGER addIndex(12)

kennytv (approved):

PR #3850, opened 2024-05-14
Filter unknown block entity types in 1.20.3->.5

FlorianMichael:
1.20.5 stopped supporting unknown block entity types (due to the changes made to registry codecs) while 1.20.3 silently ignored invalid data, therefore checking the id range of 1.20.3 block entity types. 

Closes https://github.com/ViaVersion/ViaVersion/issues/3804


kennytv:
Made it cleaner in https://github.com/ViaVersion/ViaVersion/commit/2d41eb52a6692e7552242091d749c3058d99f9e9

PR #3854, opened 2024-05-15
Add rewriter function for legacy chunk types to BlockRewriter

FlorianMichael:
Also, deduplicates a lot of existing code into handleChunk()


common/src/main/java/com/viaversion/viaversion/rewriter/BlockRewriter.java
```diff
@@ -173,6 +173,39 @@ public void register() {
         });
     }
 
+    public void registerLevelChunk(C packetType, Type<Chunk> chunkType, Type<Chunk> newChunkType) {
+        registerLevelChunk(packetType, chunkType, newChunkType, null);
+    }
+
+    public void registerLevelChunk(C packetType, Type<Chunk> chunkType, Type<Chunk> newChunkType, BiConsumer<UserConnection, Chunk> chunkRewriter) {
```
kennytv, line 180: nullable annotations

FlorianMichael: Fixed, thanks. also made newChunkType NotNull since it's never actually null in any implementations.

kennytv (approved):

PR #3862, opened 2024-05-18
Fixup LODESTONE_TRACKER rewriting in 1.20.3->.5 and backwards

FlorianMichael:
Makes LodestonePos and LodestoneDimension optional like in 1.20.4, I'm not sure if the handling is 100% correct (e.g. when having only LodestoneDimension 1.20.4 would still show a glint while 1.20.6 now wouldn't) but issues https://github.com/ViaVersion/ViaBackwards/issues/755 and https://github.com/ViaVersion/ViaVersion/issues/3861 should be fixed now.


FlorianMichael:
Maybe you could also look at the handling so we don't get things wrong again?

kennytv (approved):
Should be fine, the other minor parsing differences we can't really handle without changing other behavior

Barvalg:
Platform: 1.20.6--82--0ad09de%20%28MC%3A%201.20.6%29 </br>ViaVersion (5.0.0-SNAPSHOT): Even with dev-Even%20with%20dev-brightgreen?style=flat-square) </br>ViaBackwards(5.0.0-SNAPSHOT): Even with dev-Even%20with%20dev-brightgreen?style=flat-square) </br>

kennytv:
Not fixable

PR #3866, opened 2024-05-19
Fix FIREWORKS rewriting in 1.20.5->.3

FlorianMichael:
Fixes flight duration having the wrong type and also adds duration 1 as default value, it looks like 1.20.6 stopped sending default values to the client like 1.20.4 (so we will need to add them manually for every item where it matters).

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
@@ -387,7 +387,7 @@ public final class BlockItemPacketRewriter1_20_5 extends ItemRewriter<Clientboun
             return dataItem;
         }
 
-        if (dataConverter.isBackupInconvertibleData()) {
+        if (dataConverter.backupInconvertibleData()) {
             // In 1.20.6, some items have default values which are not written into the components
             if (item.identifier() == 1105 && !data.contains(StructuredDataKey.FIREWORKS)) {
                 data.set(StructuredDataKey.FIREWORKS, new Fireworks(1, new FireworkExplosion[0]));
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
```diff
@@ -387,6 +387,11 @@ public Item toOldItem(final UserConnection connection, final Item item, final St
             return dataItem;
         }
 
+        // In 1.20.6, some items have default values which are not written into the components
+        if (item.identifier() == 1105 && !data.contains(StructuredDataKey.FIREWORKS)) {
```
kennytv, line 391: only if backupInconvertibleData is set in the dataConverter

kennytv (changes requested):

common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
FlorianMichael: Maybe we should consider renaming the boolean then?

FlorianMichael: Anyway, fixed

common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/StructuredDataConverter.java
```diff
@@ -871,6 +873,10 @@ private <T> void register(final StructuredDataKey<T> key, final SimpleDataConver
         rewriters.put(key, c);
     }
 
+    public boolean isBackupInconvertibleData() {
```
kennytv, line 876: no is prefix

FlorianMichael: Fixed

kennytv (approved):

PR #3868, opened 2024-05-19
Fixup ComponentUtil#legacyToJsonString behavior with styles

FlorianMichael:
Looking at the git history and the old code it seems that this was missed previously, item name in 1.12->1.13 and item lore in 1.13->1.14 should both be italic default (I tested this using native server and client versions).


Kichura (approved):

kennytv (changes requested):
This will break items with color formatting

FlorianMichael:
Did some testing with various items with custom name and lore on a 1.12 server with both 1.12 and 1.16 client and they look equal now, tested items:

minecraft:fishing_rod 1 0 {display:{Lore:["§aHello","§b§kWorld"]}}
minecraft:fishing_rod 1 0 {display:{Lore:["Test"]}}
minecraft:stone 1 0 {display:{Name:"Test"}}
minecraft:stone 1 0 {display:{Name:"§aTest"}}

kennytv (approved):

PR #3874, opened 2024-05-25
Remove outdated dependencies/plugins from plugin.yml

FlorianMichael:
PS support got moved to https://github.com/ViaVersionAddons/ProtocolSupportCompat, so I believe this can be removed. Not sure if we could remove other dependencies as well like **PacketListenerApi**

### DIFF
--- a/bukkit/src/main/resources/plugin.yml
+++ b/bukkit/src/main/resources/plugin.yml
@@ -6,7 +6,7 @@ description: ${description}
 api-version: 1.13
 folia-supported: true
 loadbefore: [ ProtocolLib, ProxyPipe, SpigotLib, SkinRestorer ]
-softdepend: [ ProtocolSupport, PacketListenerApi ]
+softdepend: [ PacketListenerApi ]
 commands:
   viaversion:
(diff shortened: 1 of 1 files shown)


bukkit/src/main/resources/plugin.yml
```diff
@@ -6,7 +6,7 @@ description: ${description}
 api-version: 1.13
 folia-supported: true
 loadbefore: [ ProtocolLib, ProxyPipe, SpigotLib, SkinRestorer ]
```
Kichura, line 8: "SkinsRestorer" is spelled wrong in `loadbefore`.

FlorianMichael: I believe it wasn't when this got added, but seems like they changed it to SkinsRestorer

kennytv:
Every single other one except for ProtocolLib can also be removed

FlorianMichael:
Thought so, removed

Kichura (approved):

kennytv (approved):

PR #3880, opened 2024-05-26
Fix recipe rewriting in 1.20.3->.5

FlorianMichael:
This is needed in the transition protocol between DataItem and StructuredItem because types of Item have changed and would throw ArrayStoreException otherwise.

Closes https://github.com/ViaVersion/ViaFabricPlus/issues/412 and can be tested on both servers mentioned there.


kennytv:
Proper fix is in https://github.com/ViaVersion/ViaVersion/commit/e26f63c7034314c994551c6c0c25bcc99d0fd47c

PR #3889, opened 2024-05-28
Handle items with invalid amount in 1.20.3->.5

FlorianMichael:
1.20.4 can send items which have a bigger amount than vanilla usually accepts, and the client doesn't validate it, 1.20.5 on the other hand checks against MAX_STACK_SIZE, meaning we should set MAX_STACK_SIZE to always have the item amount to emulate old behavior. Inventory actions are broken in both versions after sending an iron sword with a stack size of 64, so /shrug about that.

Fixes https://github.com/ViaVersion/ViaVersion/issues/3882

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
@@ -457,4 +457,11 @@ public interface ViaVersionConfig extends Config {
      * @return the value sent to 1.19+ clients on join
      */
     boolean enforceSecureChat();
+
+    /**
+     * Handles items with invalid count values (higher than max stack size) on 1.20.3 servers.
+     *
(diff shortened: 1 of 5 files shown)


AoElite:
Tested it & it does appear to fix https://github.com/ViaVersion/ViaVersion/issues/3882

RaphiMC:
Won't this cause items to be unmergable clientside? If the player has two half stacks of stone (32) and tries to merge them into one stack (64) won't that cause issues if the client thinks that the stack is already full (because 32 is the max stack size of each)

kennytv (changes requested):
Needs a config option, defaulting to false

FlorianMichael:
@RaphiMC right, the fix should only apply if the item amount is larger than the correct max stack size

FlorianMichael:
Added a setting for it as well as proper validation if the fix is really needed

PR #3890, opened 2024-05-28
Move default values into handle method (1.20.3->.5)

FlorianMichael:
Shouldn't be inside of toOldItem() since it's not part of the actual component<->nbt conversion, also prevents confusing reference to backupInconvertibleData() which could now be removed?


FlorianMichael:
Oh right, but why isn't it in ViaBackwards then

PR #3898, opened 2024-06-01
Remove bungee options from config.yml

FlorianMichael:
Another pull request will add them into ViaBungee to prevent bloating the config.yml with platform settings which aren't considered as mainly supported anymore.


kennytv (changes requested):
Missing lots of references and loading

FlorianMichael:
Removed

kennytv (approved):

PR #3900, opened 2024-06-01
Improve some grammar and rename update checker setting

FlorianMichael:


common/src/main/java/com/viaversion/viaversion/configuration/AbstractViaConfig.java
```diff
@@ -103,7 +103,7 @@ public void reload() {
     }
 
     protected void loadFields() {
-        checkForUpdates = getBoolean("checkforupdates", true);
```
kennytv, line 106: Needs handling of the old field

kennytv (changes requested):

common/src/main/java/com/viaversion/viaversion/configuration/AbstractViaConfig.java
FlorianMichael: Done

FlorianMichael:
Removed handling for the old value again so we don't need to do major changes to the config system just for it.

kennytv (approved):

PR #3901, opened 2024-06-02
Restructure build-system to only publish important/usable modules

FlorianMichael:
There isn't really much a point to publishing either template or compat submodules to the repository server as they shouldn't be depended on anyway, I think we should also consider not publishing some of the platform impls as well, fabric/velocity, for example, can't and shouldn't be used at all. (Not sure about bukkit). I decided to come up with this because we now have platforms which are external (sponge/bungee) and changed artifacts, making it not really possible to publish them at all (At least not in the classic way with com.viaversion.viaversion-<name>).

Not sure if extensions#publishJavaComponents should be kept or not, can be removed if you want.

If we agree on something here, I'll open a PR to ViaBackwards as well and do the same for ViaRewind.

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -34,5 +34,3 @@ dependencies {
 java {
     withJavadocJar()
 }
-
-publishShadowJar()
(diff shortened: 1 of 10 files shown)


kennytv:
Yes, publishJavaComponents can be removed. the other platform modules should stay published

FlorianMichael:
Done, it should now publish everything except compat/template modules.

FlorianMichael:
I think we can move publishShadowJar() into the shadow conventions now, should this be done?

kennytv:
Yup, just inline it there

FlorianMichael:
Done on both PRs

kennytv (approved):

PR #3912, opened 2024-06-04
Add EntityRewriter#registerBlockStateHandler

FlorianMichael:
Deduplicates some common code by moving out minecart block state rewriting into an own handler like other common metadata rewriters

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_13_2to1_14/rewriter/EntityPacketRewriter1_14.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_13_2to1_14/rewriter/EntityPacketRewriter1_14.java
@@ -260,9 +260,11 @@ public class EntityPacketRewriter1_14 extends EntityRewriter<ClientboundPackets1
     protected void registerRewrites() {
         filter().mapDataType(Types1_14.ENTITY_DATA_TYPES::byId);
         registerEntityDataTypeHandler(Types1_14.ENTITY_DATA_TYPES.itemType, Types1_14.ENTITY_DATA_TYPES.optionalBlockStateType, Types1_14.ENTITY_DATA_TYPES.particleType);
-        registerBlockStateHandler(EntityTypes1_14.ABSTRACT_MINECART, 10);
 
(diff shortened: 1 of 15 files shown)


common/src/main/java/com/viaversion/viaversion/rewriter/EntityRewriter.java
```diff
@@ -223,6 +223,17 @@ public void registerEntityDataTypeHandler(@Nullable EntityDataType itemType, @Nu
         });
     }
 
+    public void registerMinecartBlockStateHandler(final EntityType entityType) {
+        registerMinecartBlockStateHandler(entityType, 11);
+    }
```
kennytv, line 228: No 11 default, even though it didn't change for a few versions it's still too volatile of a random value

FlorianMichael: Looking at the usages it's clearly the majority of protocols and there is no hint the index will change, but I'll remove it and rename the method then since it's not bounded to minecarts anymore?

FlorianMichael:
Updated both PRs

common/src/main/java/com/viaversion/viaversion/protocols/v1_13_2to1_14/rewriter/EntityPacketRewriter1_14.java
```diff
@@ -260,6 +260,7 @@ public void register() {
     protected void registerRewrites() {
         filter().mapDataType(Types1_14.ENTITY_DATA_TYPES::byId);
         registerEntityDataTypeHandler(Types1_14.ENTITY_DATA_TYPES.itemType, Types1_14.ENTITY_DATA_TYPES.optionalBlockStateType, Types1_14.ENTITY_DATA_TYPES.particleType);
+        registerBlockStateHandler(EntityTypes1_14.ABSTRACT_MINECART, 10);
 
         filter().type(EntityTypes1_14.ENTITY).addIndex(6);
```
kennytv, line 265: And this

kennytv (approved):

PR #3932, opened 2024-06-12
Add /viaversion player <name|*> command

FlorianMichael:
Adding a command to get the version of a singleton player, I think this can be useful for debugging since the list commands tends to be unreadable on servers with lots of players.

image


kennytv:
Maybe playerinfo/player instead of view?

kennytv (approved):

PR #3933, opened 2024-06-12
Cleanup /viaversion list and pps command

FlorianMichael:
I don't see any need for getOnlinePlayers() as we can directly pull connections from the connection manager which also tracks clients not translated by Via*, also cleaning up commands.


kennytv:
Platform online players can be different to currently cached connections, the method in platform should stay, even if unused with the other changes

kennytv (approved):

PR #3941, opened 2024-06-15
Add method to remove sub commands from command handler

FlorianMichael:
Adds API so ViaLoader and ViaFabricPlus can properly unload sub commands which are not supported or working.


kennytv:
Removing it by the string key would be better

FlorianMichael:
Yup

kennytv (approved):

Issue #3944, opened 2024-06-16
Login error with Hardcore Mode enabled


LeonTG:
### `/viaversion dump` Output

https://dump.viaversion.com/55f1e7cda473148cc991af4588190c56ee5f63981ee60a122f0c25b14fcfbada

### Console Error

https://mclo.gs/7nd25xw

### Bug Description

Since the hardcore boolean is encoded into the gamemode id byte (1.8 NMS: https://gyazo.com/8ee87d97ce5bf5d657ba113878f736fb, the `b` boolean is if its hardcore mode) and the ViaVersion rewriter does not take this into account (ViaVersion gamemode parsing code: https://gyazo.com/e14070fdf56f4cc66ae8ee2be74314fc), it errors when the gamemode id isn't recognized, in my case it returns value 10, which ViaVersion now doesn't know what to do with

Before 5.0.0 it would just return null, but not it throws an error instead

### Steps to Reproduce

1. Login to a 1.8 server that has hardcore mode enabled
2. You should not be able to join, you get the error I provided and get kicked

### Expected Behavior

I expect to be able to login without any errors and not get kicked.

### Checklist

- [X] Via plugins are only running on **EITHER** the backend servers (e.g. Paper) **OR** the proxy (e.g. Velocity), **not on both**.
- [X] I have included a ViaVersion dump.
- [X] If applicable, I have included a paste (**not a screenshot**) of the error.
- [X] I have tried the latest build(s) from https://ci.viaversion.com/ and the issue still persists.

Barvalg:
![Platform: git--Arctic--%2207d11b5%22%20%28MC%3A%201.8.8%29](https://img.shields.io/badge/Platform-git--Arctic--%2207d11b5%22%20%28MC%3A%201.8.8%29-3498db?style=flat-square) </br>![ViaVersion (5.0.0): Even with master](https://img.shields.io/badge/ViaVersion%20(5.0.0)-Even%20with%20master-brightgreen?style=flat-square) </br>

LeonTG:
I cloned the repo and added this to the ViaVersion gamemode class to solve the issue on my end
![image](https://github.com/ViaVersion/ViaVersion/assets/8997490/2e6bf7e9-5a21-4ad1-ad65-a758b41e8982), but I figure ViaVersion wants a more proper fix than what I have done

kennytv:
That looks fine, you can PR it if you want (though maybe as &= 7)

LeonTG:
[...]

PR #3946, opened 2024-06-16
Fix players crashing when logging into a 1.8 server on hardcore mode

LeonTG:
This fixes the issue #3944 that I opened

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/minecraft/GameMode.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/minecraft/GameMode.java
@@ -40,8 +40,6 @@ public enum GameMode {
     }
 
     public static GameMode getById(int id) {
-        id &= -9; // remove the hardcore mode flag
-
         return switch (id) {
             case 0 -> SURVIVAL;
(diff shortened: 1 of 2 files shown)


FlorianMichael (changes requested):
The hardcore bit flag should be removed there https://github.com/ViaVersion/ViaVersion/blob/b367caa4b2c5d1079de973074cdb5e976378cabd/common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/rewriter/PlayerPacketRewriter1_9.java#L199 and not inside the enum, only the LOGIN packet is doing it.

LeonTG:
> The hardcore bit flag should be removed there
> 
> https://github.com/ViaVersion/ViaVersion/blob/b367caa4b2c5d1079de973074cdb5e976378cabd/common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/rewriter/PlayerPacketRewriter1_9.java#L199
> 
> and not inside the enum, only the LOGIN packet is doing it.

pushed that change now

FlorianMichael (approved):
Thank you!

PR #3953, opened 2024-06-17
Add handleMenuType function to ItemRewriter

FlorianMichael:
Allows VB (and other protocols in the future) to rewrite both window title and type at the same time.


common/src/main/java/com/viaversion/viaversion/rewriter/ItemRewriter.java
```diff
@@ -112,20 +112,22 @@ public void registerOpenScreen(C packetType) {
...
-                    }
-
-                    wrapper.write(Types.VAR_INT, mappedId);
-                });
+                handler(wrapper -> handleWindowType(wrapper));
             }
         });
     }
 
+    public void handleWindowType(final PacketWrapper wrapper) {
```
kennytv, line 120: menu type

kennytv (approved):

PR #3969, opened 2024-06-22
Rename ComponentRewriter->TextRewriter

FlorianMichael:
Now that item components with 1.20.5+ are a thing the name component rewriter became even more unclear than before, TextRewriter better reflections what the rewriter actually does.


FlorianMichael:
After internal team discussion we decided against this.

PR #3997, opened 2024-07-03
Fix: delay chat acknowledgements instead of spoofing for 1.20.3 / 1.20.5 translation

### LINKED ISSUE #3940: Failed to validate message acknowledgements
### `/viaversion dump` Output

https://dump.viaversion.com/c7556e94f9c820f8698f959afe6f3965e7b69b5085992208fb6d0d1ec12662ee

### Console Error

https://mclo.gs/7mePk6U

### Bug Description

Players with old client version are getting sometimes a "Failed to validate message acknowledgements from" error. This happened after the Update to 1.20.6 and the latest ViaVersion version. 
Could be related to this https://www.spigotmc.org/resources/viaversion.19254/update?update=72944

### Steps to Reproduce

Join the server 1.20.6 with older client version. 
Send messages, it is random there is no pattern. However, one of the player think that it happens mostly when you try to chat after running any command

### Expected Behavior

player send message without getting kicked

### Additional Server Info

we use ViaVersion and ViaBackwards.
Tested with latest version of ViaVersion

### Checklist

- [X] Via plugins are only running on **EITHER** the backend servers (e.g. Paper) **OR** the proxy (e.g. BungeeCord), **not on both**.
- [X] I have included a ViaVersion dump.
- [X] If applicable, I have included a paste (**not a screenshot**) of the error.
- [X] I have tried the latest build(s) from https://ci.viaversion.com/ and the issue still persists.

Gegy:
To translate between a 1.20.5 client and 1.20.3 server, command packets without a last seen update need to attach a spoofed last seen update. Currently, this assumes that all messages were acknowledged by the client - which is not always the case.

An easy reproduction case is to mute a player using the Social Interactions screen, let them chat, and then send a message. As ViaVersion produces its own acknowledgements, this now does not match with what the client produced and signed its messages with - so the signature will fail verification.

As an alternative, we always pass through the same acknowledgement bit set - and for 'spoofed' messages, we reuse the last one that we received. In the case of the chat ack packet with just an offset, we need to hold back the window by at least 20 messages to ensure we don't start making claims about whether the client saw a message that ViaVersion cannot yet make a judgement on.

Closes #3940. This is a counterpart to https://github.com/PaperMC/Velocity/pull/1366, implementing a similar strategy with delayed message acknowledgements.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/Protocol1_20_3To1_20_5.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/Protocol1_20_3To1_20_5.java
@@ -56,6 +56,7 @@ import com.viaversion.viaversion.rewriter.SoundRewriter;
 import com.viaversion.viaversion.rewriter.StatisticsRewriter;
 import com.viaversion.viaversion.rewriter.TagRewriter;
 import com.viaversion.viaversion.util.ProtocolLogger;
+import java.util.BitSet;
 import java.util.UUID;
 
 import static com.viaversion.viaversion.util.ProtocolUtil.packetTypeMap;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/Protocol1_20_3To1_20_5.java
```diff
@@ -241,16 +230,19 @@ protected void registerPackets() {
         cancelServerbound(ServerboundPackets1_20_5.DEBUG_SAMPLE_SUBSCRIPTION);
     }
 
-    private void replaceChatAck(final PacketWrapper wrapper, final AcknowledgedMessagesStorage storage) {
-        wrapper.read(Types.VAR_INT); // Offset
-        wrapper.read(Types.ACKNOWLEDGED_BIT_SET); // Acknowledged
-        writeChatAck(wrapper, storage);
+    private void fixChatAck(final PacketWrapper wrapper, final AcknowledgedMessagesStorage storage) {
+        final int offset = wrapper.read(Types.VAR_INT); // Offset
+        final BitSet acknowledged = wrapper.read(Types.ACKNOWLEDGED_BIT_SET); // Acknowledged
```
kennytv, line 235: No need for the name comments here

common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/Protocol1_20_3To1_20_5.java
```diff
@@ -188,7 +169,16 @@ protected void registerPackets() {
                 }
             }
 
-            replaceChatAck(wrapper, storage);
+            fixChatAck(wrapper, storage);
+        });
+        registerServerbound(ServerboundPackets1_20_5.CHAT_ACK, ServerboundPackets1_20_3.CHAT_ACK, wrapper -> {
```
kennytv, line 174:
```suggestion
        registerServerbound(ServerboundPackets1_20_5.CHAT_ACK, wrapper -> {
```
If the name is the same, they can get automatically mapped

kennytv (approved):
Looks good otherwise, thanks!

common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/Protocol1_20_3To1_20_5.java
Gegy: Neat, thanks! :)

PR #4003, opened 2024-07-04
Only rewrite entity data if entity is present

FlorianMichael:
Minecraft 1.8-1.21 checks if the entity exists when handling the entity data packet, this causes Via to rewrite entity data for non-existing entities where the entity type is null.

Proper fix for https://github.com/ViaVersion/ViaRewind/issues/527


kennytv (changes requested):
You need to track all entity types in all protocols for this, which currently is not the case

PR #4006, opened 2024-07-06
Change Attack Speed for 1.8 -> 1.9

lowercasebtw:
The previous attack speed value would still show a slight cooldown, the new value removes the cooldown entirely.

The bug that was mentioned in the comment above the code doesn't seem to happen in modern version (1.21) from testing.
I wasn't able to reproduce the bug in 1.9 also.

The reason the value is 23 is because I tested ingame using a sword until the cooldown dissapeared.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/rewriter/EntityPacketRewriter1_9.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/rewriter/EntityPacketRewriter1_9.java
@@ -303,7 +303,7 @@ public class EntityPacketRewriter1_9 extends EntityRewriter<ClientboundPackets1_
                         properties.put(key, new Pair<>(value, modifiers));
                     }
                     
-                    properties.put("generic.attackSpeed", new Pair<>(23, ImmutableList.of( // Neutralize modifiers
+                    properties.put("generic.attackSpeed", new Pair<>(23.0, ImmutableList.of( // Neutralize modifiers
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/rewriter/EntityPacketRewriter1_9.java
```diff
@@ -302,11 +302,8 @@ public void register() {
                         }
                         properties.put(key, new Pair<>(value, modifiers));
                     }
-
-                    // == Why 15.9? ==
-                    // Higher values hides the cooldown but it bugs visual animation on hand
-                    // when removing item from hand with inventory gui
-                    properties.put("generic.attackSpeed", new Pair<>(15.9, ImmutableList.of( // Neutralize modifiers
+                    
+                    properties.put("generic.attackSpeed", new Pair<>(23.0, ImmutableList.of( // Neutralize modifiers
```
FlorianMichael, line 306:
```suggestion
                    properties.put("generic.attackSpeed", new Pair<>(20.0, ImmutableList.of( // Neutralize modifiers
```

Seems to be enough to remove the cooldown animation

lowercasebtw: Were you holding a sword when testing? Cause the cooldown indicator didn’t disappear till I got to 23

FlorianMichael (approved):
Thank you!

PR #4010, opened 2024-07-07
Add several fixes using item data components

RaphiMC:
This PR adds several fixes for older protocols using the new item data components.

This first fix adds instant sword blocking on 1.8 servers (It doesn't rely on inventory tracking and thus always works and has zero delay). The consume time has been set to the same value a 1.8 client uses (3600 seconds). The fix has been tested on a 1.8 server and works with or without the old shield blocking fix.

The second fix changes the crossbow max damage on <= 1.17.1 servers to fix desynced visuals on newer clients.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
@@ -27,7 +27,6 @@ import com.viaversion.nbt.tag.NumberTag;
 import com.viaversion.nbt.tag.StringTag;
 import com.viaversion.nbt.tag.Tag;
 import com.viaversion.viaversion.api.Via;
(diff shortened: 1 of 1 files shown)


FlorianMichael (approved):
We have the same code working in VFP/ViaProxy for a long time now without issues

common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
```diff
@@ -609,6 +612,19 @@ public Item toStructuredItem(final UserConnection connection, final Item old) {
         return item;
     }
 
+    private void appendItemDataFixComponents(final UserConnection connection, final Item item) {
+        if (connection.getProtocolInfo().serverProtocolVersion().olderThanOrEqualTo(ProtocolVersion.v1_17_1)) {
```
kennytv, line 616: put the server protocol into a variable

kennytv (commented):
Both should get a comment explaining why

kennytv (approved):

PR #4025, opened 2024-07-19
Fix block predicate conversion in 1.20.3->1.20.5

FlorianMichael:
No idea what we did their but these are blocks and not items lol


kennytv (changes requested):
You first need to map the id to 1.20.3

kennytv (approved):

PR #4038, opened 2024-07-22
Add ViaManager#addPostEnableListener

FlorianMichael:
Adds API so platforms can execute code after the platform implementation has executed their loading code (for example to override providers).

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/ViaManager.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/ViaManager.java
@@ -141,6 +141,13 @@ public interface ViaManager {
     void addEnableListener(Runnable runnable);
 
     /**
+     * Adds a runnable to be executed when ViaVersion has finished its init after the full server load.
+     *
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/viaversion/viaversion/api/ViaManager.java
```diff
@@ -140,6 +140,13 @@ default void setDebug(boolean debug) {
      */
     void addEnableListener(Runnable runnable);
 
+    /**
+     * Adds a runnable to be executed when ViaVersion has finished its init after the full server load.
+     *
+     * @param runnable runnable to be executed
+     */
+    void addLateEnableListener(Runnable runnable);
```
kennytv, line 148: PostEnable

FlorianMichael: Renamed, I also removed the javadoc and added @ Override since the methods come from the interface

kennytv (approved):

PR #4054, opened 2024-07-28
Add config option for truncating 1.11+ chat messages

calcastor:
This PR adds a config option for truncating 1.11+ chat. 

With this option disabled, 1.11+ clients can send longer chat messages on legacy servers which modify `PacketPlayInChat`, without needing to use ViaChatFixer.

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
@@ -464,4 +464,11 @@ public interface ViaVersionConfig extends Config {
      * @return true if enabled
      */
     boolean cancelBlockSounds();
+
+    /**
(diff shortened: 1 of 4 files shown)


FlorianMichael (changes requested):
Thanks for your contribution!
In my opinion, this should either completely handle messages longer than 100 characters (by for example splitting the chat message packet) or not be done at all since the current code would still require the server to make changes and could confuse users to think the setting automatically handles everything. (Also plugins can easily override protocol handlers like ViaChatFixer is doing at https://github.com/ViaVersionAddons/ViaChatFixer/blob/089f6f9cc070654299efad9c9c95836ab725c460/src/main/java/fr/mrmicky/viachatfixer/ChatHandler.java#L42 so a plugin doing this could just keep the modification in its own code rather than this setting which would require the user to manually disable)

calcastor:
Ideally I would change this to split messages only if necessary (as in, if the server is not modified) but this isn't something I have any ideas as to how to do. Choosing not to go forward at the moment.

PR #4060, opened 2024-07-29
Fix enchantment emulation in 1.20.5->1.21 on non Paper servers

FlorianMichael:
Moves out Paper specific code so the fix can also run on non-Paper server and also uses older/deprecated API to run on legacy versions like 1.8-1.12. Note that this still doesn't support most enchantments on older/non-Paper servers, but the most important enchantment is supported now.

Closes https://github.com/ViaVersion/ViaVersion/issues/3951


bukkit/src/main/java/com/viaversion/viaversion/bukkit/listeners/v1_20_5to1_21/PlayerChangeItemListener.java
```diff
@@ -23,58 +23,39 @@
...
-public final class PlayerChangeItemListener extends ViaBukkitListener {
+public class PlayerChangeItemListener extends ViaBukkitListener {
 
-    private final Enchantment efficiency = Enchantment.getByKey(NamespacedKey.minecraft("efficiency"));
-    private final Enchantment aquaAffinity = Enchantment.getByKey(NamespacedKey.minecraft("aqua_affinity"));
-    private final Enchantment depthStrider = Enchantment.getByKey(NamespacedKey.minecraft("depth_strider"));
-    private final Enchantment soulSpeed = Enchantment.getByKey(NamespacedKey.minecraft("soul_speed"));
-    private final Enchantment swiftSneak = Enchantment.getByKey(NamespacedKey.minecraft("swift_sneak"));
+    // Use legacy function and names here to support all versions
+    private final Enchantment efficiency = Enchantment.getByName("DIG_SPEED");
```
kennytv, line 41: I want these to work with Spigot's trash bytecode editing disabled, so it needs to first try the new name and then fall back to the old

kennytv (changes requested):

kennytv (approved):

PR #4064, opened 2024-08-02
Handle pitch/yaw->data item frame rotation change in 1.16.4->1.17

FlorianMichael:
See https://github.com/ViaVersion/ViaBackwards/pull/848


kennytv:
What are example values of this resulting in what issue that is fixed by this? In my testing the pitch and yaw are handled fine without the server sending any rotation extra there

kennytv (approved):
Nevermind looks good

PR #4069, opened 2024-08-04
Move bukkit-legacy into bukkit

FlorianMichael:
I don't see any reason for keeping bukkit-legacy as the only relevant change is getTypeId() which can be accessed through getType().getId(). Also renames some remaining packages to the new protocol name format.


kennytv:
Eventually these methods will be gone, keeping a separate module is going to be easier than using reflection

PR #4073, opened 2024-08-05
Fix typo in ViaManagerImpl#destroy log message

FlorianMichael:


kennytv:
The updated message makes less sense

kennytv (approved):

PR #4082, opened 2024-08-08
Update 1.21 version range

RaphiMC:


FlorianMichael (changes requested):
gradle.properties needs updating and the project version needs to be updated too.

FlorianMichael (approved):

kennytv (approved):

Issue #3643, opened 2024-01-12
Placing blocks is insanely broken (found the issue)


michaello01:
### `/viaversion dump` Output

https://dump.viaversion.com/f6140354455ea1f92a6e64ac2b1ef4a35eed5771445e2bebf5fc9c57853b9526

### Bug Description

When you place a block, it gets reverted to air, then after 1 or 2 tick it is being shown as the proper block. It makes placing blocks, jumping on them really difficult and chaotic. Sing editing is still broken.
After a really long investigation I noticed what's going on. I increased this delay: https://github.com/ViaVersion/ViaVersion/blob/aecef3c3f8db69b26617c9d4a0615e52a5bf0b2f/bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitAckSequenceProvider.java#L41
to 20, and sign editing works, after 20 ticks the sign disappear if the placement was blocked, but when during these 20 ticks you place a block, then you place a sign, you don't wait 20 ticks to sign to get refreshed, but less. That means ack tasks are not related directly  to the action. I also felt that block placing is better with longer delay. With 20 ticks delay block placing feels great, with smaller delay like 2 or 3 ticks when you place blocks really fast, for example when you place the blocks that touches other blocks that you just placed, the placing gets blocked and it starts to get really chaotic. Can you check that please? I spent many, many hours investigating that. 
Disabling the ack provider or making it run immediately doesn't help.
Increasing the delay makes the canceled placements stay for longer - for example you can levitate on blocks. Block change packet doesn't update the placement, the only thing that comes to my mind is to look into the ack system and handle it differently. This is a massive bug and it breaks the block placement completely.

### Steps to Reproduce

1. Login with 1.19+ client on 1.8 server.
2. Place some blocks.
3. You will notice that air appears right after the block is placed, then it goes back to the normal block.
4. Sign editing doesn't work.

### Expected Behavior

Smooth block placement like on Hypixel for example and like it was before. Working sign editing, but it is small issue compared to placing blocks.

### Additional Server Info

BungeeCord, ViaVersion running on backend servers.

### Checklist

- [X] Via plugins are only running on **EITHER** the backend servers (e.g. Paper) **OR** the proxy (e.g. BungeeCord), **not on both**.
- [X] I have included a ViaVersion dump.
- [X] If applicable, I have included a paste (**not a screenshot**) of the error.
[...]

PR #4093, opened 2024-08-13
Properly handle block change acknowledgements in 1.18.2 -> 1.19

RaphiMC:
Recoded block change acknowledgement handling to fix block placement and block breaking lag issues.

Comparison video with 1 second simulated ping: https://upload.raphimc.net/file/0ZtebB8Se2NJrSQN/adx9mAEt2pcj7ytc/idea64_ZBLOxYOdHa.mp4
Left is with my patch and right is without my patch (Both running connected to a 1.8 server over ViaProxy)

I left the bukkit code mostly untouched. Might be worth to consider removing the bukkit code because the protocol level fix is almost perfect and the bukkit code has its own issues (#3643)


michaello01:
I think with these changes and commented this line
https://github.com/ViaVersion/ViaVersion/blob/dea6dee7f9cf22634d062f452c511e8a091233ed/bukkit/src/main/java/com/viaversion/viaversion/bukkit/platform/BukkitViaLoader.java#L180
it works smooth and without any problems, tested on bukkit platform. :)

ImAwsumm (approved):

PR #4097, opened 2024-08-14
Setup dependabot for GH Actions and add workflow for updating Gradle

FlorianMichael:
Prevents manual/unregular commits of updating both GH actions and the Gradle wrapper, sadly dependabot doesn't support Gradle Version Catalogs so it can't be used for dependencies.


Kichura (approved):
lgtm.

kennytv (approved):

PR #4106, opened 2024-08-18
Refactor around base protocol implementations for ViaVersion addons

FlorianMichael:
This PR renames/refactors both base protocols for client and serverbound packets and exposes more code to prevent ViaLegacy (and future ViaVersion addons) copying pasting base protocol code into their project.

The `InitialBaseProtocol` (previously `BaseProtocol`) now only contains the serverbound handshake handler (assuming the handshake packet will never change) and is the root/inital base protocol which is always set.

Both `ServerboundBaseProtocol1_7` and `ClientboundBaseProtocol1_7` are now only loaded for 1.7+ clients and contain more methods to be used by ViaLegacy (and other addons).

Furthermore the `ProtocolManager#getBaseProtocol` function has been renamed and now supports multiple base protocols for the same server version, since any old usage will now most likely break I removed the old function.


RaphiMC:
Tested it with ViaLegacy and ViaBedrock. Both still work with the changes

kennytv (approved):

PR #4112, opened 2024-08-21
Allow protocols to handle client intention packet

FlorianMichael:
This is currently not possible without manually added base protocols which is the case in both ViaLegacy/ViaBedrock. This way we can get rid of it completely: https://github.com/ViaVersion/ViaLegacy/blob/main/src/main/java/net/raphimc/vialegacy/api/protocol/PreNettyBaseProtocol.java

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/base/InitialBaseProtocol.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/base/InitialBaseProtocol.java
@@ -129,8 +129,7 @@ public class InitialBaseProtocol extends AbstractProtocol<BaseClientboundPacket,
                 protocols.remove(this);
                 wrapper.apply(Direction.SERVERBOUND, State.HANDSHAKE, protocols);
             } catch (CancelException e) {
-                wrapper.user().setActive(false);
-                return;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/base/InitialBaseProtocol.java
```diff
@@ -123,6 +123,16 @@ protected void registerPackets() {
                 wrapper.set(Types.VAR_INT, 0, serverProtocol.getOriginalVersion());
             }
 
+            // Send client intention into the pipeline in case protocols down the line need to transform it
+            try {
+                final List<Protocol> protocols = new ArrayList<>(pipeline.pipes());
+                protocols.remove(this);
+                wrapper.apply(Direction.SERVERBOUND, State.HANDSHAKE, protocols);
+            } catch (CancelException e) {
+                wrapper.user().setActive(false);
```
kennytv, line 132: Should maybe just ignore cancellation?

FlorianMichael: I looked at the if (versionProvider == null) code

kennytv (approved):

PR #4113, opened 2024-08-21
Cleanup LOGIN/STATUS packet handlers 

FlorianMichael:
Some combinations with the override boolean for ViaFabricPlus and other addons


FlorianMichael:
I only added those which I would use but I can remove the ones you don't like.

common/src/main/java/com/viaversion/viaversion/protocols/v1_12_2to1_13/Protocol1_12_2To1_13.java
```diff
@@ -173,11 +173,11 @@ protected void registerPackets() {
 
         WorldPacketRewriter1_13.register(this);
 
-        registerClientbound(State.LOGIN, ClientboundLoginPackets.LOGIN_DISCONNECT.getId(), ClientboundLoginPackets.LOGIN_DISCONNECT.getId(), wrapper -> {
+        registerClientbound(State.LOGIN, ClientboundLoginPackets.LOGIN_DISCONNECT.getId(), wrapper -> {
```
kennytv, line 176: all of these can already just be replaced with
```java
registerClientbound(State.LOGIN, ClientboundLoginPackets.LOGIN_DISCONNECT, wrapper -> {
```
etc.

kennytv (changes requested):
Actually, none of these are "standard" or needed for the main projects, helper methods are better added in the projects that need them

FlorianMichael:
I'll keep those extra methods in ViaFabricPlus then; Resolved the suggestion

kennytv (approved):

PR #4118, opened 2024-08-22
Split base protocols list

RaphiMC:
Splits the base protocol list in clientbound and serverbound. Useful when client and server version are not both 1.7+ (ViaLegacy, ViaBedrock)
Related to https://github.com/ViaVersion/ViaVersion/pull/4106

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
@@ -29,6 +29,7 @@ import com.viaversion.viaversion.api.protocol.ProtocolManager;
 import com.viaversion.viaversion.api.protocol.ProtocolPathEntry;
 import com.viaversion.viaversion.api.protocol.ProtocolPathKey;
 import com.viaversion.viaversion.api.protocol.packet.ClientboundPacketType;
+import com.viaversion.viaversion.api.protocol.packet.Direction;
 import com.viaversion.viaversion.api.protocol.packet.PacketType;
 import com.viaversion.viaversion.api.protocol.packet.PacketWrapper;
 import com.viaversion.viaversion.api.protocol.packet.ServerboundPacketType;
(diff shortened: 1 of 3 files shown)


FlorianMichael (approved):

common/src/main/java/com/viaversion/viaversion/protocol/ProtocolManagerImpl.java
```diff
@@ -365,11 +371,20 @@ > VersionedPacketTransformer<C, S> createPacketTransformer(ProtocolVersion input
     }
 
     @Override
-    public List<Protocol> getBaseProtocols(ProtocolVersion serverVersion) {
+    public List<Protocol> getBaseProtocols(@Nullable ProtocolVersion clientVersion, @Nullable ProtocolVersion serverVersion) {
```
kennytv, line 374: Any reason for them being nullable?

FlorianMichael: At least the `clientboundBaseProtocolVersion` in InitialBaseProtocol will be null if no valid base protocol can be found

kennytv: Is that actually a valid client state?

FlorianMichael: Yes, special versions not using the RedirectProtocolVersion can't have the base protocol set via the registry (as they might compare equal to normal versions). In this case, they always need to manually add the base protocol.

xism4: > clientboundBaseProtocolVersion

I don't see any referceny that says its a state. Others like handshake, login, ping etc are

FlorianMichael: This is not related to protocol states but just to whether it's legitimate to not have a clientbound base protocol found in the registry.

kennytv (approved):

PR #4122, opened 2024-08-24
Add option to hide scoreboard numbers

Pablete1234:
Useful for minigame servers where scoreboard numbers are always used purely for ordering purposes. Enabling the config makes it so 1.20.3+ clients will see the scoreboard without the red numbers on the right.

Tested on a 1.8 server with 1.21.1 client, both with the setting off and on, produces the expected results


FlorianMichael (approved):
Works and looks good.

kennytv (approved):

PR #4136, opened 2024-09-04
Rewrite display name component in player info packet in 1.20.3->.5

FlorianMichael:
Closes https://github.com/ViaVersion/ViaFabricPlus/issues/529. Decided to keep this only in this very protocol since joining this server worked with older ViaFabricPlus versions for MC 1.20.4.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/rewriter/ComponentRewriter.java
+++ b/common/src/main/java/com/viaversion/viaversion/rewriter/ComponentRewriter.java
@@ -40,6 +40,7 @@ import com.viaversion.viaversion.util.ComponentUtil;
 import com.viaversion.viaversion.util.SerializerVersion;
 import com.viaversion.viaversion.util.TagUtil;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import java.util.BitSet;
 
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/viaversion/viaversion/rewriter/ComponentRewriter.java
```diff
@@ -139,6 +140,42 @@ public void registerTabList(final C packetType) {
         });
     }
 
+    public void registerPlayerInfoUpdate1_20_3(final C packetType) {
+        protocol.registerClientbound(packetType, wrapper -> {
+            final BitSet actions = wrapper.passthrough(Types.PROFILE_ACTIONS_ENUM);
```
Camotoy, line 145: Maybe add a `if (!actions.get(5)) return` here since you're not doing anything else with the packet? (Then you can also remove the check below)

FlorianMichael: There can be multiple actions at once and we need to passthrough all bytes

kennytv (approved):

PR #4139, opened 2024-09-05
Fix invalid book view not visible in 1.20.3->.5

FlorianMichael:
Adds back the old "Invalid book tag" view which mc 1.20.4 and older had when the book compound tag was invalid, the logic is 1:1 with how the vanilla client detects it.

image


common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
```diff
@@ -1163,47 +1163,63 @@ private void updateWritableBookPages(final StructuredDataContainer data, final C
...
+        final String author = tag.getString("author");
         final ListTag<StringTag> pagesTag = tag.getListTag("pages", StringTag.class);
-        final CompoundTag filteredPagesTag = tag.getCompoundTag("filtered_pages");
-        if (pagesTag == null) {
-            return;
+
+        boolean valid = author != null && title != null && title.length() <= 32 && pagesTag != null;
+        if (valid) {
+            for (final StringTag page : pagesTag) {
+                if (page.getValue().length() > 32767) {
```
kennytv, line 1173:
```suggestion
                if (page.getValue().length() > Short.MAX_VALUE) {
```

kennytv (approved):

PR #4143, opened 2024-09-12
Add EntityTrackerBase#hasClientEntityId and don't use -1 as not set indicator

FlorianMichael:
-1 can be used as valid entity id and should not be used as placeholder for not set.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/data/entity/EntityTrackerBase.java
+++ b/common/src/main/java/com/viaversion/viaversion/data/entity/EntityTrackerBase.java
@@ -37,7 +37,7 @@ public class EntityTrackerBase implements EntityTracker, ClientEntityIdChangeLis
     protected final Int2ObjectMap<TrackedEntity> entities = new Int2ObjectOpenHashMap<>();
     private final UserConnection connection;
     private final EntityType playerType;
-    private int clientEntityId = -1;
+    private Integer clientEntityId;
     private int currentWorldSectionHeight = -1;
     private int currentMinY;
     private String currentWorld;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/viaversion/viaversion/data/entity/EntityTrackerBase.java
```diff
@@ -102,15 +102,15 @@ public void clearEntities() {
     }
 
     @Override
-    public int clientEntityId() {
+    public Integer clientEntityId() {
```
kennytv, line 105: I'd prefer still returning a primitive here and instead adding a hasClientEntityId method

FlorianMichael: Because of backwards compatibility? Then I still have to return -1 and it generally is more likely to be used wrong as people would always need to remember wrapping it around hasClientEntityId.

RaphiMC: It could throw an exception if being called when the client entity id not set to avoid using the method wrongly

FlorianMichael: Good point, removed

kennytv: Most packets will also throw on the client if join game hasn't been sent yet, or it'll be Via failing to track it. So hasClientEntityId won't always be necessary, but throwing is necessary

kennytv (approved):

PR #4156, opened 2024-09-20
Fix enchantment glint behaviour in 1.10->1.11

FlorianMichael:


common/src/main/java/com/viaversion/viaversion/protocols/v1_10to1_11/rewriter/ItemPacketRewriter1_11.java
```diff
@@ -63,10 +80,15 @@ public Item handleItemToServer(UserConnection connection, Item item) {
...
-            if (item.tag().isEmpty()) {
-                item.setTag(null);
+        CompoundTag tag = item.tag();
+        if (tag != null) {
+            if (tag.contains(nbtTagName())) {
+                item.setAmount(tag.<ByteTag>removeUnchecked(nbtTagName()).asByte());
+                if (tag.isEmpty()) {
+                    item.setTag(null);
+                }
+            } else if (tag.remove(nbtTagName("clearEnch")) != null) {
```
kennytv, line 90: Shouldnt this be in its own if outside the else?

kennytv (approved):

PR #4170, opened 2024-10-08
Write correct on ground state in 1.20.5->1.21 placement rotation fix

FlorianMichael:


Cyramek:
I don't think this option should work for 1.8 servers, as you can use it as regen cheat, because of way 1.8 server calculates how many ticks passed based on client movement packets

RaphiMC:
> I don't think this option should work for 1.8 servers, as you can use it as regen cheat, because of way 1.8 server calculates how many ticks passed based on client movement packets

ViaVersion didn't cancel the additional movement packet for older client versions where the client sent it itself anyway, so this fix is not worse than spamming right click with a <=1.20.5 client

kennytv (approved):

PR #4180, opened 2024-10-12
Change description of dump sub command to be more generic

FlorianMichael:
Since VV also runs on clients or as standalone proxy where the command system is loaded as well


Kichura (approved):

kennytv:
"platform information" or something

FlorianMichael:
Yup

kennytv (approved):

PR #4189, opened 2024-10-18
Add option to construct ChunkSectionLightImpl without block light

RaphiMC:
Adds an optional boolean to the ChunkSectionLightImpl constructor to not add block light by default. This makes it a lot easier to use this class for the modern Minecraft lighting engine where both lights are optional


FlorianMichael (approved):

kennytv:
Merged to preview manually

PR #4202, opened 2024-10-26
Fix and add missing javadocs in MappingData interface

FlorianMichael:


api/src/main/java/com/viaversion/viaversion/api/data/MappingData.java
```diff
@@ -35,56 +35,82 @@ public interface MappingData {
...
+     * Returns the mapped attribute id, or 0 if unmapped.
+     *
+     * @param id unmapped attribute id
+     * @return mapped attribute id, or 0 if unmapped
+     * @throws NullPointerException if this mappingdata does not hold attribute mappings
+     */
     int getNewAttributeId(int id);
 
+    /**
+     * Returns the mapped entity id, or 0 if unmapped.
```
kennytv, line 100: "entity id"

FlorianMichael: Fixed

kennytv (approved):

PR #4208, opened 2024-10-28
Fix item_name<->default name priority for items in 1.21->1.21.2

FlorianMichael:
Minecraft 1.21.1 client:
custom_name > item_name > default/generated (e.g. player heads/potions)

Minecraft 1.21.2 client:
custom name > default/generated > item_name


kennytv (changes requested):
This should only be done for the items that need it

FlorianMichael:
Why?

kennytv:
It's unnecessary messing with item data and extra work when only like 3 types of items actually need it. The custom name is used for a lot of other things on the client

FlorianMichael:
Makes sense, done

kennytv (approved):

PR #4211, opened 2024-10-29
Add gradle task to quickly test/debug code changes

RaphiMC:
Adds a gradle task which starts ViaProxy with the current code changes. This allows quick and easy testing of new changes without complex server/client mod setup.

Also allows hotswapping if added as IntelliJ run configuration:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,3 +110,6 @@ nb-configuration.xml
 
 ### MacOS ###
 .DS_Store
+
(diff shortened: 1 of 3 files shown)


FlorianMichael (approved):
Sounds fine to me given that ViaProxy is now an official ViaVersion software.

kennytv (approved):

PR #4220, opened 2024-10-30
Drop idle packets where only horizontal collision changed in 1.21->1.21.2

FlorianMichael:
Since Minecraft 1.21.2, the client now also sends idle packets when the horizontal collision flag changed. This can cause bad packets on older servers where the flag didn't exist (causing the client to send idle packets without the ground state changing).


RaphiMC (changes requested):
The tracked flags should be reset when the player is respawning

kennytv (approved):

PR #4223, opened 2024-10-30
Change the consume animation of swords to block

RaphiMC:
Restores the blocking animation for swords on 1.8 or lower servers

image

This only works on 22w44a+ client due to a client bug. The animation exists in 1.21.3, but doesn't render in first person.


common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/BlockItemPacketRewriter1_21_2.java
```diff
@@ -486,7 +487,16 @@ public static void updateItemData(final Item item) {
                 final Consumable1_21_2.ApplyStatusEffects applyStatusEffects = new Consumable1_21_2.ApplyStatusEffects(new PotionEffect[]{effect.effect()}, effect.probability());
                 consumeEffects[i] = new Consumable1_21_2.ConsumeEffect<>(0 /* add status effect */, Consumable1_21_2.ApplyStatusEffects.TYPE, applyStatusEffects);
             }
-            dataContainer.set(StructuredDataKey.CONSUMABLE1_21_2, new Consumable1_21_2(food.eatSeconds(), 1 /* eat */, sound, true, consumeEffects));
+
+            int animation = 1; // eat
+            if (user.getProtocolInfo().serverProtocolVersion().olderThanOrEqualTo(ProtocolVersion.v1_8)) {
```
FlorianMichael, line 492: This should be moved to the handleItemToClient function since updateItemData is also called in ViaBackwards

FlorianMichael (approved):

kennytv (approved):

PR #4228, opened 2024-10-31
Fix protocol error when an empty lock is sent (picking beacon, etc.)

valaphee:
In 1.8, when picking any container block, with 1.21.3 the client will disconnect with the following protocol error:
```
Caused by: io.netty.handler.codec.DecoderException: Failed to decode: Failed to parse either. First: Failed to parse either. First: Not a string: null; Second: Not a json array: null; Second: Not a JSON object: null missed input: {custom_name:""} {components:{custom_name:""}}
```


kennytv:
If it only happens with 1.21.3 clients, it should go into the `LOCK` handling in `BlockItemPacketRewriter1_21_2`

valaphee:
I'm blind, thanks

FlorianMichael (approved):
Thanks, I changed the code a bit to match the styling of other rewriters.

valaphee:
It also doesn't accept spaces

PR #4238, opened 2024-11-03
Fix sound id rewriting in instrument components

FlorianMichael:


kennytv:
Instrument ids did not change

PR #4245, opened 2024-11-05
Changed "fix-1_21-placement-rotation" to true by default

Pro4D:
Changes the config option "fix-1_21-placement-rotation" to true by default in order to restore vanilla behavior.


Kichura:
> Changes the config option "fix-1_21-placement-rotation" to true by default in order to restore vanilla behavior.

While this does restore vanilla behavior, this is false on purpose because of anti-cheat plugins. therefore this PR is concerning.

Pro4D:
Are there any explicit examples of what problems this would cause for anti-cheat plugins? And which ones?

Firecrest-1337 (approved):
I’ve tested this on my own server and it seems to work fine, the only issue i see is that it may accidentally flag anti cheat but definitely not in any major way, especially considering they usually get flagged anyway for unrelated things. I think this should get implemented because liquid behaviour is really really important for pvp.

FlorianMichael (approved):
Sounds reasonable

PR #4276, opened 2024-11-21
Add config option for 1.20.4->1.20.5 sword blocking fix

FlorianMichael:
Some people might want to disable it since it breaks the sword blocking animation for clients older than 1.21.4


FlorianMichael:
Yeah, I couldn't think of a name, thanks

kennytv (approved):

PR #4282, opened 2024-11-25
Fix block ids not getting remapped

valaphee:
Block ids are not getting remapped, because `getBlockMappings()` is always null as the mappings are not loaded before the registration/initialization.

It's also possible to load the mappings before and wait for all mappings to be loaded (as this is done on an executor thread), but this approach is way less invasive, and still correct.

This only affects players using Geyser at the moment (as Geyser uses the block id to translate block event, chest animation not working etc.), but who knows if Java Edition extends its validation, or alters behavior.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/rewriter/BlockRewriter.java
+++ b/common/src/main/java/com/viaversion/viaversion/rewriter/BlockRewriter.java
@@ -67,11 +67,11 @@ public class BlockRewriter<C extends ClientboundPacketType> {
     }
 
     public void registerBlockEvent(C packetType) {
-        if (protocol.getMappingData().getBlockMappings() == null) {
-            return;
-        }
-
         protocol.registerClientbound(packetType, wrapper -> {
+            if (protocol.getMappingData().getBlockMappings() == null) {
+                return;
+            }
+
             wrapper.passthrough(positionType); // Location
             wrapper.passthrough(Types.UNSIGNED_BYTE); // Action id
             wrapper.passthrough(Types.UNSIGNED_BYTE); // Action param


kennytv:
Those methods shouldn't be called early or if the protocol has no mappings as it'd otherwise suggest bad calls, geyser should add null checks where appropriate instead

common/src/main/java/com/viaversion/viaversion/rewriter/BlockRewriter.java
```diff
@@ -67,10 +67,6 @@ public static <C extends ClientboundPacketType> BlockRewriter<C> for1_20_2(final
     }
 
     public void registerBlockEvent(C packetType) {
-        if (protocol.getMappingData().getBlockMappings() == null) {
-            return;
-        }
-
         protocol.registerClientbound(packetType, wrapper -> {
             wrapper.passthrough(positionType); // Location
```
kennytv, line 71: The null check and return should be in here

api/src/main/java/com/viaversion/viaversion/api/data/MappingDataBase.java
```diff
@@ -174,12 +174,12 @@ public int getNewBlockStateId(final int id) {
 
     @Override
     public int getNewBlockId(final int id) {
-        return checkValidity(id, blockMappings.getNewId(id), "block");
+        return blockMappings != null ? checkValidity(id, blockMappings.getNewId(id), "block") : id;
```
kennytv, line 177: And these two diffs dropped

kennytv (changes requested):

common/src/main/java/com/viaversion/viaversion/rewriter/BlockRewriter.java
```diff
@@ -67,15 +67,14 @@ public static <C extends ClientboundPacketType> BlockRewriter<C> for1_20_2(final
...
             wrapper.passthrough(Types.UNSIGNED_BYTE); // Action id
             wrapper.passthrough(Types.UNSIGNED_BYTE); // Action param
             final int blockId = wrapper.passthrough(Types.VAR_INT);
+            if (protocol.getMappingData().getBlockMappings() == null) {
```
FlorianMichael, line 75: You can move the check up before all `passthrough` calls I believe.

valaphee: didn't know thx

kennytv (approved):

PR #4284, opened 2024-11-26
1.12: Recipe placeholders are stone

valaphee:
Minecraft 1.13 introduced custom recipes to the recipe book, before that, numerical ids were assigned to recipes.
When sending 1.12 recipes, null is used as ingredient, but this causes the 1.21.2 translation to replace it with stone.

This PR strips out null ingredients, and uses empty ingredients in the 1.12.2->1.13 recipe map.


kennytv:
The rest looks good - but how did recipes even work for you without https://github.com/ViaVersion/ViaVersion/commit/896c6accfbe9f7cecba0f5bd303a3f02073c148c 👀

valaphee:
> The rest looks good - but how did ghost recipes/client recipe ingredient checks even work for you without [896c6ac](https://github.com/ViaVersion/ViaVersion/commit/896c6accfbe9f7cecba0f5bd303a3f02073c148c) 👀

Forgot to include it in this PR ^^'

common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/RecipeRewriter1_20_5.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    protected void handleIngredient(final PacketWrapper wrapper) {
+        final Item[] items = wrapper.read(itemArrayType());
+        final List<Item> newItems = new ArrayList<>(items.length);
+        for (final Item item : items) {
+            if (item == null || item.isEmpty()) continue;
+            newItems.add(item);
+        }
+        wrapper.write(mappedItemArrayType(), newItems.toArray(items));
```
kennytv, line 43: You have to pass new String[0] in toArray

kennytv (changes requested):

common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/RecipeRewriter1_20_5.java
valaphee: was already in the process, you are quick ^^'

kennytv (approved):
Looks good, thank you!

PR #4287, opened 2024-11-27
Fix footstep particle causing null kick

lowercasebtw:
Added checks for when getMappingData returns null and getParticleMappings returns null.

Also swapped the equals check around to stop null issues.

This fixes the recent bug I found when playing 1.8.x bedwarspractice and was getting kicked when using invis and the footstep particles spawned.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/ParticleRewriter1_21_2.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/ParticleRewriter1_21_2.java
@@ -18,7 +18,6 @@
 package com.viaversion.viaversion.protocols.v1_21to1_21_2.rewriter;
 
 import com.viaversion.viaversion.api.connection.UserConnection;
-import com.viaversion.viaversion.api.data.ParticleMappings;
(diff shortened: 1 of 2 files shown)


lowercasebtw:
Fixed

common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/ParticleRewriter1_21_2.java
```diff
@@ -43,12 +44,13 @@ private void floatsToARGB(final Particle particle, final int fromIndex) {
     @Override
     public void rewriteParticle(final UserConnection connection, final Particle particle) {
         super.rewriteParticle(connection, particle);
-
-        final String identifier = protocol.getMappingData().getParticleMappings().mappedIdentifier(particle.id());
-        if (identifier.equals("minecraft:dust_color_transition")) {
+        
+        ParticleMappings particleMappings = protocol.getMappingData().getParticleMappings();
```
FlorianMichael, line 48: Revert this change/diff again, no need for the extra field

FlorianMichael (changes requested):

lowercasebtw:
Done

kennytv (approved):

PR #4290, opened 2024-11-29
Send full position update packet in 1.20.5->1.21 placement rotation fix

FlorianMichael:
Seems to be better after talking with anti cheat developers


RaphiMC (changes requested):
Don't track the position in clientbound packets. This will cause a race condition

RaphiMC (approved):

kennytv (approved):

Envarad:
is there a way to fix the position packet from 1.20.4 to 1.21.1 for via proxy?

FlorianMichael:
This fix is included in the next ViaProxy release or in dev builds at https://ci.viaversion.com/view/Platforms/job/ViaProxy/

Envarad:
It doesn't appear to be working, which is why I'm asking. Anytime i click or hold right click, i appear to be lagged in place. Minecraft 1.20.4 sends a position packet right before interacting with the item/block which the server doesn't expected and thus the anticheat flags it. 1.20.6 changed that behaviour. therefore, is there an actual fix to this issue or do i have to suffer through the lagging out whenever i wanna place blocks/water, webs, etc

RaphiMC:
> It doesn't appear to be working, which is why I'm asking. Anytime i click or hold right click, i appear to be lagged in place. Minecraft 1.20.4 sends a position packet right before interacting with the item/block which the server doesn't expected and thus the anticheat flags it. 1.20.6 changed that behaviour. therefore, is there an actual fix to this issue or do i have to suffer through the lagging out whenever i wanna place blocks/water, webs, etc

This PR is for the other direction where the packet has to be emulated. Your issue would belong to ViaBackwards. For further discussion please use the discord channels

PR #4292, opened 2024-11-30
fix: ViaBackwards 923

404Setup:
ViaBackwards [#923](https://github.com/ViaVersion/ViaBackwards/issues/923)


kennytv (changes requested):
That won't work, our schedulers are explicitly anything but synchronized to the server

404Setup:
So should it run in Folia's GlobalRegion?

That's what I think.

bukkit/src/main/java/com/viaversion/viaversion/bukkit/platform/FoliaViaTask.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.google.common.base.Preconditions;
+import com.viaversion.viaversion.api.platform.PlatformTask;
+import io.papermc.paper.threadedregions.scheduler.ScheduledTask;
+
+public record FoliaViaTask(ScheduledTask task) implements PlatformTask<ScheduledTask> {
+
+    @Override
+    public void cancel() {
+        Preconditions.checkArgument(task != null, "Task cannot be cancelled");
```
kennytv, line 28: The returned task is notnull, no need for the check

bukkit/src/main/java/com/viaversion/viaversion/ViaVersionPlugin.java
```diff
@@ -75,6 +76,14 @@ public ViaVersionPlugin() {
         conf.reload();
     }
 
+    /**
+     * @deprecated use {@link Via#getAPI()} instead
+     */
+    @Deprecated(forRemoval = true)
+    public static ViaVersionPlugin getInstance() {
+        return instance;
+    }
```
kennytv, line 85: Don't move this

404Setup: In my memory I didn't do this, maybe my IDE did it automatically. 

Already rolled back the operation.

kennytv (approved):
Thanks

PR #4296, opened 2024-12-06
Fix slot index check

riku6460:


kennytv (approved):
that's what happens when I push a fix the day of, thanks

PR #4299, opened 2024-12-06
Sword blocking: consumables for 1.21.4+, back to shields for 1.20.5-1.21.3

roccodev:
This PR has two commits, the first reimplements consumable blocking for 1.21.4 only, and removes all blocking-related shields from 1.21.4 clients, using consumables instead.

The reason consumable blocking is removed for 1.20.5-1.21.3 is that it currently doesn't work properly. With both `shield-blocking` and `sword-blocking-via-consumable` enabled (the former is required to still let 1.9-1.20.4 clients block), blocking in 1.20.5-1.21.3 *seems* to be working on the client, but the block doesn't actually get registered on the server *most of the time*. 

What I believe is happening is that the client starts eating, then when it receives a shield it stops using the item so the block gets cancelled server side. (I checked with a debugger and ViaVersion's `setBlocking` is correctly called with 1 `true` and 1 `false` for each shield cycle)

Obviously another solution would be to remove shields from 1.20.5-1.21.3 clients, but given the lack of first-person blocking animation for consumables, it could be somewhat confusing for users. If servers wish to have instant blocking on those versions, they can use the instant shield blocking config option, which works perfectly fine in my experience.

This change only affects 1.8 servers, and I'd probably expect most if not all of them to only care about supporting the latest version. So it shouldn't be that big of an issue to not support consumable blocking for versions in-between.

The other commit removes the `sword-blocking-via-consumable` config option. After the changes, the only reason to keep the option is to move back to shields in 1.21.4+, something you'd probably only do for consistency's sake.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/storage/EntityTracker1_9.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/storage/EntityTracker1_9.java
@@ -103,8 +103,7 @@ public class EntityTracker1_9 extends EntityTrackerBase {
      * The item in the offhand will be cleared if there is no sword in the main hand.
      */
     public void syncShieldWithSword() {
-        if (Via.getConfig().swordBlockingViaConsumable()
-                && user().getProtocolInfo().protocolVersion().newerThanOrEqualTo(ProtocolVersion.v1_21_4)) {
+        if (user().getProtocolInfo().protocolVersion().newerThanOrEqualTo(ProtocolVersion.v1_21_4)) {
             // If sword blocking is done through consumables, don't add a shield.
             return;
         }
@@ -194,8 +193,8 @@ public class EntityTracker1_9 extends EntityTrackerBase {
                         if ((data & 0x10) == 0x10) {
                             // If sword blocking is done through consumables, don't add a shield. We can use
                             // 1.20.5 here because consumables display properly in 3rd person even in 1.20.5-1.21.3
-                            if (validBlocking.contains(entityId) && (!Via.getConfig().swordBlockingViaConsumable()
-                                    || user().getProtocolInfo().protocolVersion().olderThan(ProtocolVersion.v1_20_5))) {
+                            if (validBlocking.contains(entityId) &&
+                                    user().getProtocolInfo().protocolVersion().olderThan(ProtocolVersion.v1_20_5)) {
                                 Item shield = new DataItem(442, (byte) 1, (short) 0, null);
                                 setSecondHand(entityId, shield);
                             } else {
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/storage/EntityTracker1_9.java
```diff
@@ -185,7 +191,10 @@ public void handleEntityData(int entityId, List<EntityData> entityDataList) {
                     byte data = (byte) entityData.getValue();
                     if (entityId != getProvidedEntityId() && Via.getConfig().isShieldBlocking()) {
                         if ((data & 0x10) == 0x10) {
-                            if (validBlocking.contains(entityId)) {
+                            // If sword blocking is done through consumables, don't add a shield. We can use
+                            // 1.20.5 here because consumables display properly in 3rd person even in 1.20.5-1.21.3
+                            if (validBlocking.contains(entityId) &&
+                                    user().getProtocolInfo().protocolVersion().olderThan(ProtocolVersion.v1_20_5)) {
```
RaphiMC, line 197: Why are you checking for 1.20.5 here. You removed the 1.20.5 consumable blocking fix, so you would need the old fix for 1.20.5 as well

roccodev: You're right, I think this was leftovers from some experiments I did earlier. Indeed, without the change 1.20.5-1.21.3 clients can't see shields on other players. 

Thanks!

common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/storage/EntityTracker1_9.java
```diff
@@ -185,7 +191,9 @@ public void handleEntityData(int entityId, List<EntityData> entityDataList) {
                     byte data = (byte) entityData.getValue();
                     if (entityId != getProvidedEntityId() && Via.getConfig().isShieldBlocking()) {
                         if ((data & 0x10) == 0x10) {
-                            if (validBlocking.contains(entityId)) {
+                            // If sword blocking is done through consumables, don't add a shield.
+                            if (validBlocking.contains(entityId) &&
+                                    user().getProtocolInfo().protocolVersion().olderThan(ProtocolVersion.v1_21_4)) {
```
RaphiMC, line 196: Move the protocol version check up to the ``isShieldBlocking`` if or is there a specific reason to put it here?

roccodev: I think that can work. Before I also had to check for `swordBlockingViaConsumable`, so it made sense to support shields that were present before a config reload

RaphiMC (approved):

FlorianMichael (approved):

PR #4312, opened 2024-12-14
Fix entity defaults across all protocols

FlorianMichael:
Minecraft <= 1.13.2 silently ignores unknown entity type ids and doesn't print any warning, while we currently still track them. Minecraft 1.14+ will spawn a pig if the entity type is invalid while we track ENTITY as wrong type. Both cases can cause various entity data issues where the server sends wrong data for non-existing entities while we still handle it.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/rewriter/EntityRewriter.java
+++ b/common/src/main/java/com/viaversion/viaversion/rewriter/EntityRewriter.java
@@ -510,6 +510,9 @@ public abstract class EntityRewriter<C extends ClientboundPacketType, T extends 
         }
 
         final EntityType entityType = typeFromId(trackMappedType ? mappedTypeId : typeId);
+        if (entityType == null) {
+            return null;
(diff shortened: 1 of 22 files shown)


common/src/main/java/com/viaversion/viaversion/rewriter/EntityRewriter.java
```diff
@@ -510,6 +510,9 @@ public EntityType trackAndRewrite(final PacketWrapper wrapper, final int typeId,
         }
 
         final EntityType entityType = typeFromId(trackMappedType ? mappedTypeId : typeId);
+        if (entityType == null) {
+            return null;
+        }
         tracker(wrapper.user()).addEntity(entityId, entityType);
```
kennytv, line 516: Doesn't matter too much, but me want empty lines after braces if a "normal" statement of a "next" logical step follows

kennytv (approved):

PR #4315, opened 2024-12-15
Do not persist credentials

Kichura:
Yoinked from [EnZaXD's BaseProject #26](https://github.com/FlorianMichael/BaseProject/pull/26).

### DIFF
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -10,5 +10,7 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - name: Update Gradle Wrapper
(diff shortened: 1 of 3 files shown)


.github/workflows/update-gradle-wrapper.yml
```diff
@@ -10,5 +10,7 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
```
kennytv, line 14: I don't think this will work

kennytv (commented):
Why is that needed?

.github/workflows/update-gradle-wrapper.yml
Kichura: This actually does work and has been tested before it was ever submitted, and it was caught by the `zizmor` rust library.

kennytv (approved):

PR #4338, opened 2025-01-03
Fix CONTAINER transition handling in 1.20.3->1.20.5

FlorianMichael:
Fixes shulker box contents not being visible/broken and also disconnects caused by chat components.

Should also fix the issue mentioned in https://github.com/ViaVersion/ViaBackwards/issues/888


common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
```diff
@@ -1486,8 +1487,16 @@ private void updateBlockEntityTag(final UserConnection connection, @Nullable fin
                 data.set(StructuredDataKey.BASE_COLOR, baseColorIntTag.asInt());
             }
 
-            if (tag.contains("Items")) {
-                updateItemList(connection, data, tag, "Items", StructuredDataKey.CONTAINER1_20_5, true);
+            final ListTag<CompoundTag> itemsTag = tag.getListTag("Items", CompoundTag.class);
+            if (itemsTag != null) {
+                final Item[] items = new Item[27];
```
kennytv, line 1492: `Items` isn't just used for shulker boxes so you cannot presize like that, and otherwise would still need slot checks below

kennytv (changes requested):

common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
```diff
@@ -1486,8 +1487,38 @@ private void updateBlockEntityTag(final UserConnection connection, @Nullable fin
                 data.set(StructuredDataKey.BASE_COLOR, baseColorIntTag.asInt());
             }
 
-            if (tag.contains("Items")) {
-                updateItemList(connection, data, tag, "Items", StructuredDataKey.CONTAINER1_20_5, true);
+            final ListTag<CompoundTag> itemsTag = tag.getListTag("Items", CompoundTag.class);
+            if (itemsTag != null) {
+                int highestSlot = 0;
+
+                for (int i = 0; i < Math.min(itemsTag.size(), 256); i++) {
```
kennytv, line 1494:
```suggestion
                for (int i = 0, size = Math.min(itemsTag.size(), 256); i < size; i++) {
```

common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/BlockItemPacketRewriter1_20_5.java
```diff
@@ -1486,8 +1487,38 @@ private void updateBlockEntityTag(final UserConnection connection, @Nullable fin
...
+                        continue;
+                    }
+
+                    final int slot = itemTag.getByte("Slot");
+                    if (slot > highestSlot) {
+                        highestSlot = slot;
+                    }
+                }
+
+                final Item[] filteredItems = new Item[highestSlot + 1];
```
kennytv, line 1507: This also needs to be limited

kennytv (changes requested):

kennytv (approved):

PR #4352, opened 2025-01-20
Remove newly added registry entries when already sent be the server

FlorianMichael:
Remove registry entries sent by older servers which do not exist in the current version to prevent duplicated entries when we send default data, assuming the server doesn't send them.

Closes https://github.com/ViaVersion/ViaFabricPlus/issues/683
Closes https://github.com/ViaVersion/ViaFabricPlus/issues/698

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/rewriter/RegistryDataRewriter.java
+++ b/common/src/main/java/com/viaversion/viaversion/rewriter/RegistryDataRewriter.java
@@ -79,6 +79,7 @@ public class RegistryDataRewriter {
                 if (existingKeys.contains(entry.key())) {
                     continue;
                 }
+
                 updatedEntries[index++] = entry.copy();
             }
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/viaversion/viaversion/rewriter/RegistryDataRewriter.java
```diff
@@ -65,11 +67,23 @@ public RegistryEntry[] handle(final UserConnection connection, String key, Regis
...
-            entries = Arrays.copyOf(entries, length + toAddLength);
-            for (int i = 0; i < toAddLength; i++) {
-                entries[length + i] = toAdd.get(i).copy();
+            final Set<String> toAddKeys = toAdd.stream().map(RegistryEntry::key).collect(Collectors.toSet());
+
+            final RegistryEntry[] updatedEntries = new RegistryEntry[entries.length + toAdd.size()];
+            int index = 0;
+            for (final RegistryEntry entry : entries) {
+                if (!toAddKeys.contains(entry.key())) {
+                    updatedEntries[index++] = entry.copy();
```
kennytv, line 76: The original entries don't need copying, but the server-sent entry should also be preferred over what we add

kennytv (changes requested):

common/src/main/java/com/viaversion/viaversion/rewriter/RegistryDataRewriter.java
FlorianMichael: Don't we only add registry elements that older clients don't have and thus older servers don't send? In these cases it's wrong to prefer the server-sent as a Vanilla client in that version would ignore them.

kennytv (approved):

PR #4357, opened 2025-01-24
Use "intentionally_empty" sound event for 1.8 blocking

lowercasebtw:
Uses "intentionally_empty" sound event which is no sound, for the consumable blocking in 1.21.4+ on 1.8 server.


FlorianMichael (approved):
Thanks

PR #4370, opened 2025-01-31
Make functions optional which are not always required in ViaPlatform

FlorianMichael:


api/src/main/java/com/viaversion/viaversion/api/platform/ViaPlatform.java
```diff
@@ -125,15 +125,19 @@ default boolean isProxy() {
...
+    }
 
     /**
      * Send a message to a player
      *
      * @param uuid    The player's UUID
      * @param message The message to send
      */
-    void sendMessage(UUID uuid, String message);
+    default void sendMessage(UUID uuid, String message) {
```
kennytv, line 139: Ideally this too

api/src/main/java/com/viaversion/viaversion/api/platform/ViaPlatform.java
```diff
@@ -142,7 +146,9 @@ default boolean isProxy() {
      * @param message The message to kick them with
      * @return True if it was successful
      */
-    boolean kickPlayer(UUID uuid, String message);
+    default boolean kickPlayer(UUID uuid, String message) {
```
kennytv, line 149: This should not be default and at the very least handle the client if matching

kennytv (approved):

PR #4412, opened 2025-03-21
Implement pick-block for legacy servers

Pablete1234:
Currently pick-block only works in 1.12 and higher, as older bukkit versions don't have the needed methods to make it work.

This PR adds a legacy workaround using nms, aimed at 1.8 servers. It uses nms' `Block.i` method that returns an ItemStack (mostly giving the block + data in its item form), which works well for most blocks, and then relies on a fallback of `getDropType` which fills in the gap for blocks that don't seem to be mapped to an item. This last one was needed for things like sugarcane, redstone (and repeaters), doors, and some other outliers.

Items i've found not to work too well:
- Cake doesn't seem to pick-block at all (i guess the cake was a lie after all)
- Top slabs give the item without stripping the top/bottom nature, so a spruce top slab is wooden_slab:9 (instead of being wooden_slab:1), which client thinks is oak (unknown item with the :9 metadata) but when placed down is spruce again.
- Double slabs lose the type, so a spruce double_slab will just give you an oak single slab item. It's not like double slab items exist in 1.8 anyways tho
- Skulls (player heads) don't retain any metadata so they're kinda useless, always just plain skeleton skulls
- Ctrl-pick-block isn't implemented at all, so stuff like copying signs with text doesn't work

Unsure if it's worth trying to fight all the edge-cases or if we can just settle for a "works with most things" approach, which kinda seems what the 1.12 implementation did anyways (only gives the block if it's there's an item with the same id)


bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
```diff
@@ -155,7 +164,7 @@ private void spawnItem(final ItemStack item, final PlayerInventory inventory, fi
         }
 
         inventory.setItem(emptySlot, heldItem);
-        inventory.setItemInMainHand(item);
+        inventory.setItem(EquipmentSlot.HAND, item);
```
kennytv, line 167: This definitely doesn't exist in 1.8

Pablete1234: it does in sportpaper, unsure about spigot

HowardZHY: Spigot 1.8.8 only have `setItemInHand`, `setItemInMainHand` alse doesn't exists

Pablete1234: Yeah i was aware of setItemInMainHand not existing, its why i changed it, will change it further

kennytv (approved):
Thanks

Packsolite:
Seems like recent updates broke this partially for latest clients. When i pick a block using a 1.21.5 client, item data is completely lost (e.g. red wool becomes white woole, birch log becomes oak).

Pablete1234:
What server version? Consider just opening a separate issue and linking to this instead; this PR is done and closed, any follow-up will necessarily have to be done in a separated piece.

Packsolite:
> What server version? Consider just opening a separate issue and linking to this instead; this PR is done and closed, any follow-up will necessarily have to be done in a separated piece.

Server version is 1.8. Not sure if via likes issues for legacy versions like that. Also it might be the paper fork that breaks it, haven't looked into it yet...

bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
```diff
@@ -181,4 +172,54 @@ private int findEmptyHotbarSlot(final PlayerInventory inventory, final int heldS
...
+            if (PaperViaInjector.hasMethod("org.bukkit.block.data.BlockData", "getPlacementMaterial")) {
+                return (block, includeData) -> {
+                    final ItemStack item = new ItemStack(block.getBlockData().getPlacementMaterial(), 1);
+                    if (includeData && item.getItemMeta() instanceof final BlockStateMeta blockStateMeta) {
+                        blockStateMeta.setBlockState(block.getState());
+                        item.setItemMeta(blockStateMeta);
+                    }
+                    return item;
+                };
+            } else if (PaperViaInjector.hasMethod(Material.class, "isItem")) {
```
Packsolite, line 189: Okay i found the issue. Some forks have this method backported from modern versions, leading to the Legacy BlockToItem provider not being used even though it should.

Can we use a different detection here? Maybe check if the server version isLess than 1.9?

PR #4413, opened 2025-03-25
Add Copyable interface and unify copy() methods

FlorianMichael:
Adds a global Copyable interface with a copy() function used in all mutable objects.
Implements proper copy of structured item by copying all structured data elements inside the container.


api/src/main/java/com/viaversion/viaversion/api/minecraft/item/data/ChatType.java
```diff
@@ -45,7 +46,12 @@ public void writeDirect(final ByteBuf buffer, final ChatType value) {
         }
     };
 
-    public record ChatTypeDecoration(String translationKey, int[] parameters, Tag style) {
+    @Override
+    public Object copy() {
```
kennytv, line 50:
```suggestion
    public ChatType copy() {
```

PR #4449, opened 2025-04-07
Send player input every tick when in boat in 1.21->1.21.2

FlorianMichael:
Local player ticking logic:

1.21:
image

1.21.2:
image


kennytv (approved):

Kichura:
This breaks on non-spigot/paper servers, Please correct this code to avoid timing out players.

PR #4451, opened 2025-04-07
Drop (block) entity data in 1.20.3->.5 if no id is present

FlorianMichael:
@ kennytv Would appreciate if you could also verify that, but from looking at the game code I'm quite sure that the server/client need a valid id for the data to be read correctly. Meaning it's wrong on our side to try to append the id (as won't work in all cases anyway) and rather just drop data without id (as then unused on the client anyway).

Closes https://github.com/ViaVersion/ViaVersion/issues/4433


James-P-Bennett:
Absolutely agree, this is the right way to go. Since data components will probably keep changing in future Minecraft versions.

kennytv (changes requested):
We have to add ids and keep tags where possible. Old servers and often manually added data will be missing the ids but were dealt with completely fine by older clients. This applies to both entity and block entity tags

PR #4472, opened 2025-04-27
Add missing PLAYER_CHAT handlers in 1.20/1.21 versions

ImGxrke:
This PR fixes a network protocol error which occures if a signed chat message contains a show item hover event with an old custom model data.

Example log: https://mclo.gs/rlpXuL8


FlorianMichael (approved):
thanks

PR #4483, opened 2025-05-04
Make the runViaProxy task work with Gradle's configuration cache

FlorianMichael:


kennytv (changes requested):
The input jar name isn't right

kennytv (approved):

PR #4496, opened 2025-05-16
Fix legacy block picking on some paper forks

Packsolite:
As discussed in https://github.com/ViaVersion/ViaVersion/pull/4412, some forks provide the `Material#isItem()` Method for convenience, breaking the legacy support detection.

This PR changes the legacy provider to always be used when all class/method resolutions succeed during class initialization.

### DIFF
--- a/bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
+++ b/bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
@@ -18,8 +18,10 @@
 package com.viaversion.viaversion.bukkit.providers;
 
 import com.viaversion.viaversion.ViaVersionPlugin;
+import com.viaversion.viaversion.api.Via;
 import com.viaversion.viaversion.api.connection.UserConnection;
 import com.viaversion.viaversion.api.minecraft.BlockPosition;
+import com.viaversion.viaversion.api.protocol.version.ProtocolVersion;
 import com.viaversion.viaversion.bukkit.platform.PaperViaInjector;
 import com.viaversion.viaversion.bukkit.util.LegacyBlockToItem;
 import com.viaversion.viaversion.protocols.v1_21_2to1_21_4.provider.PickItemProvider;
@@ -187,14 +189,15 @@ public class BukkitPickItemProvider extends PickItemProvider {
                     }
                     return item;
                 };
-            } else if (PaperViaInjector.hasMethod(Material.class, "isItem")) {
-                return (block, includeData) -> new ItemStack(block.getType(), 1);
-            } else {
+            } else if (Via.getAPI().getServerVersion().lowestSupportedProtocolVersion().equalTo(ProtocolVersion.v1_8)) {
                 LegacyBlockToItem legacy = LegacyBlockToItem.getInstance();
                 return legacy != null ? (block, includeData) -> legacy.blockToItem(block) : (b, i) -> null;
+            } else if (PaperViaInjector.hasMethod(Material.class, "isItem")) {
+                return (block, includeData) -> new ItemStack(block.getType(), 1, (short) 0, block.getData());
+            } else {
+                return (b, i) -> null;
             }
         }
-
     }
 
     @FunctionalInterface
--- a/bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
+++ b/bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
@@ -193,7 +193,7 @@ public class BukkitPickItemProvider extends PickItemProvider {
                 LegacyBlockToItem legacy = LegacyBlockToItem.getInstance();
                 return legacy != null ? (block, includeData) -> legacy.blockToItem(block) : (b, i) -> null;
             } else if (PaperViaInjector.hasMethod(Material.class, "isItem")) {
-                return (block, includeData) -> block.getType().isItem() ? new ItemStack(block.getType(), 1, (short) 0, block.getData()) : null;
+                return (block, includeData) -> block.getType().isItem() ? new ItemStack(block.getType()) : null;
             } else {
                 return (b, i) -> null;
             }
--- a/bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
+++ b/bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
@@ -189,13 +189,23 @@ public class BukkitPickItemProvider extends PickItemProvider {
                     }
                     return item;
                 };
-            } else if (Via.getAPI().getServerVersion().lowestSupportedProtocolVersion().equalTo(ProtocolVersion.v1_8)) {
-                LegacyBlockToItem legacy = LegacyBlockToItem.getInstance();
+            }
+
+            final ProtocolVersion version = Via.getAPI().getServerVersion().lowestSupportedProtocolVersion();
+            if (version.equalTo(ProtocolVersion.v1_8)) {
+                final LegacyBlockToItem legacy = LegacyBlockToItem.getInstance();
                 return legacy != null ? (block, includeData) -> legacy.blockToItem(block) : (b, i) -> null;
             } else if (PaperViaInjector.hasMethod(Material.class, "isItem")) {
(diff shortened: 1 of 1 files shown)


FlorianMichael (approved):
Works on Spigot 1.8 & 1.12.

bukkit-legacy/src/main/java/com/viaversion/viaversion/bukkit/util/LegacyBlockToItem.java
```diff
@@ -27,20 +27,15 @@
...
         try {
-            instance = new LegacyBlockToItem();
-        } catch (ReflectiveOperationException ex) {
-            Via.getPlatform().getLogger().log(
-                Level.WARNING,
-                "Couldn't find reflection methods/fields for pick-block functionality.\n" +
-                    "Pick-block won't work on clients 1.21.4 and newer.", ex);
+            INSTANCE = new LegacyBlockToItem();
+        } catch (ReflectiveOperationException ignored) {
         }
```
kennytv, line 38: This should not silently fail if a server removes or changes any of the reflection used here, you should keep the previous code structure and only replace the isItem method check instead. like your previous suggestion for a simple version check

kennytv (changes requested):

bukkit-legacy/src/main/java/com/viaversion/viaversion/bukkit/util/LegacyBlockToItem.java
Packsolite: The reason why i decided to use a different approach is that the current implementation only supports < 1.8 and will very likely never be supported on any version ever again. So this *should* be a very safe way for checking.
But i still agree that it is kinda dirty. I will implement a version check soon.

bukkit-legacy/src/main/java/com/viaversion/viaversion/bukkit/util/LegacyBlockToItem.java
```diff
@@ -27,20 +27,15 @@
 import java.util.logging.Level;
 
 public class LegacyBlockToItem {
-    private static final LegacyBlockToItem INSTANCE;
+
     private static final Random RANDOM = new Random();
+    private static LegacyBlockToItem INSTANCE;
```
Pablete1234, line 32: I don't think there's a good reason for this to stop being final, it doesn't seem to be changed anywhere except where it's initialized

bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
```diff
@@ -187,20 +185,19 @@ static BlockToItem build() {
...
                     return item;
                 };
-            } else if (PaperViaInjector.hasMethod(Material.class, "isItem")) {
-                return (block, includeData) -> new ItemStack(block.getType(), 1);
+            } else if (LegacyBlockToItem.getInstance() != null) {
+                return (block, includeData) -> LegacyBlockToItem.getInstance().blockToItem(block);
             } else {
-                LegacyBlockToItem legacy = LegacyBlockToItem.getInstance();
-                return legacy != null ? (block, includeData) -> legacy.blockToItem(block) : (b, i) -> null;
+                return (block, includeData) -> new ItemStack(block.getType(), 1, (short) 0, block.getData());
```
Pablete1234, line 191: i guess this is a bug in my earlier pr, but this branch (which you've moved down) should be checking `block.getType().isItem()` before returning the itemstack with the blocktype, which is why it was checking for the existance of isItem to begin with

so i guess the new code should read something like:
```
else if (legacy != null):
  LegacyBlockToItem legacy = LegacyBlockToItem.getInstance();
  return (block, includeData) -> legacy.blockToItem(block);
else if (has method "isItem"):
  return (block, includeData) -> block.getType().isItem() ? new ItemStack(...) : null;
else
  return (b, i) -> null;
```
also note, you should extract legacy out of the lambda so you can save on the call each time the lambda is invoked

bukkit-legacy/src/main/java/com/viaversion/viaversion/bukkit/util/LegacyBlockToItem.java
FlorianMichael: I did that in my cleanup commit to remove the local field in the static block, but if the silent fail change has to reverted anyway, this can be final again

bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
Packsolite: Okay, please take a look at the newest version.

Packsolite:
Updated it according to all suggestions. Please take a look at it :)

Pablete1234 (approved):
Lgtm now

bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
```diff
@@ -187,14 +189,15 @@ static BlockToItem build() {
...
                     return item;
                 };
-            } else if (PaperViaInjector.hasMethod(Material.class, "isItem")) {
-                return (block, includeData) -> new ItemStack(block.getType(), 1);
-            } else {
+            } else if (Via.getAPI().getServerVersion().lowestSupportedProtocolVersion().equalTo(ProtocolVersion.v1_8)) {
                 LegacyBlockToItem legacy = LegacyBlockToItem.getInstance();
                 return legacy != null ? (block, includeData) -> legacy.blockToItem(block) : (b, i) -> null;
+            } else if (PaperViaInjector.hasMethod(Material.class, "isItem")) {
+                return (block, includeData) -> block.getType().isItem() ? new ItemStack(block.getType(), 1, (short) 0, block.getData()) : null;
```
kennytv, line 196: On 'normal' Spigot/Paper servers, this was added *after* block data was already removed. With the edited version check above, this should just use the type without data anyways

bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
```diff
@@ -187,14 +189,15 @@ static BlockToItem build() {
                     }
                     return item;
                 };
-            } else if (PaperViaInjector.hasMethod(Material.class, "isItem")) {
-                return (block, includeData) -> new ItemStack(block.getType(), 1);
-            } else {
+            } else if (Via.getAPI().getServerVersion().lowestSupportedProtocolVersion().equalTo(ProtocolVersion.v1_8)) {
```
kennytv, line 192: It works for everything until 1.13

kennytv (changes requested):

bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
Packsolite: 1.20 
image

https://helpch.at/docs/1.20/org/bukkit/block/Block.html

bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
Packsolite: The legacy data loader? No - the current implementation only works for 1.8

kennytv: That's not correct :p

bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
kennytv: Yes, it shouldn't be used after 1.12 / doesn't have to given the other version change

bukkit/src/main/java/com/viaversion/viaversion/bukkit/providers/BukkitPickItemProvider.java
kennytv: Actually I'm an idiot sorry about that

Packsolite:
Should be fine to merge finally 😅

Packsolite:
Rebased to latest.

Kichura (approved):

kennytv (approved):

PR #4498, opened 2025-05-21
Fixed rare null issues in EntityPacketRewriter1_21_2

starobot:
Got some annoying warnings because of this. Should be fixed now.


FlorianMichael:
Can you provide the server logs when the issue you are trying to fix happens? This is probably caused by something else as the storage fields should never be null inside packet handlers.

FlorianMichael:
If you can provide the error logs, please open a bug report here: https://github.com/ViaVersion/ViaVersion/issues

PR #4500, opened 2025-05-22
Don't convert creative-slot-lock/map-post-processing in 1.20.3->.5

FlorianMichael:
Also makes them not serializable in ItemDataComponentConverter. These are not supposed to be synced over the network and the item hasher will crash now in 1.21.5+ versions


common/src/main/java/com/viaversion/viaversion/rewriter/item/ItemDataComponentConverter.java
```diff
@@ -287,6 +287,12 @@ private <I> void notImplemented(final StructuredDataKey<I> key) {
         this.register(key, value -> null);
     }
 
+    private <I> void notSerializable(final StructuredDataKey<I> key) {
+        this.register(key, value -> {
+            throw new IllegalStateException(key.identifier() + " is not serializable");
```
kennytv, line 292: IllegalArgumentException

kennytv (approved):

PR #4538, opened 2025-06-21
Fix missing clouds for 1.21.5->1.21.6

booky10:
Fixes clouds missing in custom dimensions


kennytv (approved):
Not just custom dimensions - I was already wondering why I couldn't see the new clouds when I was testing VV during the snapshots :D

PR #4540, opened 2025-06-21
fix: ServerboundPackets1_21_6 documentation

dewgs:
Currently CHANGE_DIFFICULTY and CHANGE_GAME_MODE (new 1.21.6 packet) are documented as the same packet ID which makes the rest of the commented packet IDs inaccurate.

This makes absolutely no difference in ViaVersion usage, just fixes inaccurate packet ID comments.


kennytv:
Looks like the PR is coming from a weird git tree - I'll re-generate the comments within some other commit, thanks though

florianreuth:
For your next PRs you should make a new branch per pull request so you can update the master branch with new incoming commits / the git tree of every PR stays clean.

PR #4549, opened 2025-06-25
Remove team colour fix config option

FlorianMichael:
I believe this was originally added as a workaround since the original legacy string -> text component conversion was broken. I also can't find any evidence of this behaviour existing in the 1.12.2 client code, supporting that assumption. As we now use MCStructs, I believe this can be removed without any issues.

Fixes scoreboard colors being wrong on older servers.


kennytv:
This is still needed, most old servers use the prefix for coloring player names and is not related to text conversion

PR #4576, opened 2025-07-07
fix: Proxy Plugin Messaging Channel 

### LINKED ISSUE #4565: Player details plugin message only sent on first join
### `/viaversion dump` Output

https://dump.viaversion.com/e0756a6c45bad92624aa3e199169a13c56e64775d842f5deee71de3595896d57

### Server/Client Error

None

### Bug Description

When a player connects through a Velocity proxy, the **vv:proxy_details** plugin message is correctly sent to the initial backend server they join. However, when the player switches to any other backend server on the proxy, this plugin message is not sent again. This prevents other servers from receiving the player information.

### Steps to Reproduce

1. Set up a Velocity proxy with the latest version of ViaVersion.
2.  Set up two backend Paper servers. On both servers, install a simple plugin that listens for the **vv:proxy_details** plugin message and logs its contents. (Using the [example code from the documentation](https://github.com/ViaVersion/ViaVersion/wiki/Player-Details-Protocol#how-to-use-this-api-as-a-plugin-developer)).
3. Start all servers and connect to the Velocity proxy from a Minecraft client.
4. Check the console of the initial backend server. You will see the log indicating that the **vv:proxy_details** message was received successfully.
5. From in-game, switch to the second backend server (using /server <server_name>).
6. Check the console of the second backend server. No log will be printed, as the plugin message was never sent to this server.

### Expected Behavior

The **vv:proxy_details** plugin message should be sent to every backend server a player connects to, including the initial server and all subsequent servers.

### Additional Server Info

_No response_

### Checklist

criticalcynical:
This fixes the issue with proxy plugin messaging channel where the message is only sent to the first server they join due to getClientConnection returning null. This is due to it getting removed after they switch server.

This fixes #4565

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/connection/ConnectionManagerImpl.java
+++ b/common/src/main/java/com/viaversion/viaversion/connection/ConnectionManagerImpl.java
@@ -76,9 +76,9 @@ public class ConnectionManagerImpl implements ConnectionManager {
 
         UUID id = connection.getProtocolInfo().getUuid();
         if (connection.isServerSide()) {
-            serverConnections.remove(id);
+            serverConnections.remove(id, connection);
         } else {
-            clientConnections.remove(id);
+            clientConnections.remove(id, connection);
         }
 
         connection.clearStoredObjects();


florianreuth:
I think this will just never remove the connection because remove with key + value uses equals() on the current map entry which will always be false as protocols store data inside the user connection.

FlorianMichael (approved):
Actually fine I'm stupid.

florianreuth:
thank you

PR #4585, opened 2025-07-13
Update ViaNBT dependency from 5.1.0 to 5.1.2

KaiKikuchi:
As title says.

ViaVersion compiles just fine and the issue with deserializing SNBT with zero signed byte tags is fixed.


kennytv:
Note that ViaVersion itself doesn't actually use the `SNBT` class, if you want more version specific snbt deserialization that's a bit more solid you should use https://github.com/ViaVersion/mcstructs-viafier, which is also included in ViaVersion

PR #4587, opened 2025-07-13
Add cancel-swing-in-inventory option for 1.15.2->1.16 inventory swing fix

FlorianMichael:
This can cause issues with anticheat plugins and has been force disabled in VFP for a while now.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/configuration/AbstractViaConfig.java
+++ b/common/src/main/java/com/viaversion/viaversion/configuration/AbstractViaConfig.java
@@ -27,13 +27,14 @@ import com.viaversion.viaversion.protocol.BlockedProtocolVersionsImpl;
 import com.viaversion.viaversion.util.Config;
 import it.unimi.dsi.fastutil.objects.ObjectOpenHashSet;
 import it.unimi.dsi.fastutil.objects.ObjectSet;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
 import java.io.File;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


common/src/main/resources/assets/viaversion/config.yml
```diff
@@ -170,6 +170,9 @@ hide-scoreboard-numbers: false
 # Fixes 1.21+ clients on 1.20.5 servers placing water/lava buckets at the wrong location when moving fast, NOTE: This may cause issues with anti-cheat plugins.
 fix-1_21-placement-rotation: true
 #
+# If enabled, cancel swing packets sent while having an inventory opened on 1.15.2 and below servers. This can cause false positives with anti-cheat plugins.
```
Kichura, line 173: Why not make this `false` if it causes anti-cheat concerns?

FlorianMichael: Because it also fixes things / don't want to change the default value / behaviour.

common/src/main/java/com/viaversion/viaversion/configuration/AbstractViaConfig.java
```diff
@@ -27,13 +27,14 @@
 import com.viaversion.viaversion.util.Config;
 import it.unimi.dsi.fastutil.objects.ObjectOpenHashSet;
 import it.unimi.dsi.fastutil.objects.ObjectSet;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
```
kennytv, line 31: something's wrong with your setup, it's ignoring/overriding the editorconfig

kennytv (approved):

PR #4591, opened 2025-07-18
Fix NPE during 1.20.5 modifier convert

Lumine1909:
Error log: https://mclo.gs/5A5bExG

This fix is simple, and I did not do deeply research for what causes that.
Hopefully this will not causing other problems.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/ComponentRewriter1_20_5.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/ComponentRewriter1_20_5.java
@@ -1434,7 +1434,8 @@ public class ComponentRewriter1_20_5<C extends ClientboundPacketType> extends Js
     }
 
     protected ModifierData modifierDataFromTag(final CompoundTag tag) {
-        final UUID uuid = UUIDUtil.fromIntArray(tag.getIntArrayTag("uuid").getValue());
+        final IntArrayTag uuidTag = tag.getIntArrayTag("uuid");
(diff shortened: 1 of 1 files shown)


kennytv:
I can't reproduce any issues with such an item (`/give @s stick[attribute_modifiers=[{amount:0,id:"minecraft:98e5014a-fcf9-45db-98eb-11b397f012af",operation:"add_value",type:"minecraft:attack_speed"}]]`) on a 1.21.7 server with a 1.21.1 client on the latest VV and VB. an error there would mean that either the data sent is wrong or handling in Via somewhere else is wrong.

I pushed other fixes for this to VB though, please try again using the latest ci builds and otherwise send a saved hotbar from a client on the same version as the server with the item that is causing the error

Lumine1909:
> I can't reproduce any issues with such an item (`/give @s stick[attribute_modifiers=[{amount:0,id:"minecraft:98e5014a-fcf9-45db-98eb-11b397f012af",operation:"add_value",type:"minecraft:attack_speed"}]]`) on a 1.21.7 server with a 1.21.1 client on the latest VV and VB. an error there would mean that either the data sent is wrong or handling in Via somewhere else is wrong.
> 
> I pushed other fixes for this to VB though, please try again using the latest ci builds and otherwise send a saved hotbar from a client on the same version as the server with the item that is causing the error

Thank you for reply, but the issue still exist: https://mclo.gs/hsUOZvf

Also, please check log, it definitely won't happening on 1.21.1 client because the problem is during 203-205 rewriting.

kennytv:
Err sorry I meant a 1.20.1 client

Lumine1909:
I tested this issue deeply and find that it might be related to the logic that my item share plugin used to implement this feature.

I’m currently using this small fix on my server and it works good. Also, this might not happen in other servers. So I’ll temporarily close that until I find what really causes the bug.

Also, thank you for your attention to this pr.

PR #4618, opened 2025-08-07
Remove packet id comments in enums

FlorianMichael:
API methods have been using the packet type interfaces (and in the common code there enums) as a replacement for hard written packet ids for a long time now, and there is no longer a need to keep them. Especially since IDEA already shows an enum ordinal (and thus the packet id) when hovering over an entry.


kennytv:
I still use them, including from web view

PR #4641, opened 2025-09-01
Fix wrong shulker box mapping in 1.12.2->1.13

FlorianMichael:
Closes https://github.com/ViaVersion/ViaVersion/issues/4636

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_12_2to1_13/data/MappingData1_13.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_12_2to1_13/data/MappingData1_13.java
@@ -81,6 +81,14 @@ public class MappingData1_13 extends MappingDataBase {
             blockMappings.setNewId(1557, 3986); // chiseled stone bricks
         }
 
+        // Don't apply VB only mappings for purple/not-colored shulker boxes
+        // TODO Check which blocks might also be affected by inverting the VB mappings for VV
+        for (int i = 3664; i <= 3669; i++) {
+            final int index = i - 3664;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/v1_12_2to1_13/data/MappingData1_13.java
```diff
@@ -81,6 +81,14 @@ protected void loadExtras(final CompoundTag data) {
             blockMappings.setNewId(1557, 3986); // chiseled stone bricks
         }
 
+        // Don't apply VB only mappings for purple/not-colored shulker boxes
+        // TODO Check which blocks might also be affected by inverting the VB mappings for VV
```
kennytv, line 85: This isn't related to VB, it's how they were done for VV initially and mapped automatically because it *is* called purple shulker box in 1.12

common/src/main/java/com/viaversion/viaversion/protocols/v1_12_2to1_13/data/MappingData1_13.java
```diff
@@ -81,6 +81,14 @@ protected void loadExtras(final CompoundTag data) {
...
+        for (int i = 3664; i <= 3669; i++) {
+            final int index = i - 3664;
+            blockMappings.setNewId(i, 8211 + index);
+        }
+        itemMappings.setNewId(3664, 362);
```
kennytv, line 90: This should be moved to a mapping diff file if it's not a config option

kennytv (changes requested):

PR #4644, opened 2025-09-01
Rename ProtocolRunnable, always submit tick logic into event loop

FlorianMichael:


Kichura (commented):
Is this verified to not suddenly cause packet-related issues by third-party plugins / clients?

Kichura (approved):

kennytv (changes requested):

kennytv (approved):

PR #4655, opened 2025-09-27
Fix Network Protocol Error when converting wolf_variant Registry Data Packet from 1.20.6 to 1.21

amiralimollaei:
Fixes the following network protocol error in ViaFabric:

```
Description: Registry Loading

-- Head --
Thread: Render thread
Stacktrace:
	at knot//net.minecraft.registry.RegistryLoader.createLoadingException(RegistryLoader.java:280)
	at knot//net.minecraft.registry.RegistryLoader.writeAndCreateLoadingException(RegistryLoader.java:261)
	at knot//net.minecraft.registry.RegistryLoader.load(RegistryLoader.java:228)

-- Loading info --
Details:
	Errors: 
		minecraft:root/minecraft:wolf_variant: Registry must be non-empty: minecraft:wolf_variant
Stacktrace:
	at knot//net.minecraft.registry.RegistryLoader.createLoadingException(RegistryLoader.java:280)
	at knot//net.minecraft.registry.RegistryLoader.writeAndCreateLoadingException(RegistryLoader.java:261)
	at knot//net.minecraft.registry.RegistryLoader.load(RegistryLoader.java:228)
	at knot//net.minecraft.registry.RegistryLoader.loadFromNetwork(RegistryLoader.java:203)
	at knot//net.minecraft.client.network.ClientRegistries.createRegistryManager(ClientRegistries.java:95)
	at knot//net.minecraft.client.network.ClientRegistries.createRegistryManager(ClientRegistries.java:146)
	at knot//net.minecraft.client.network.ClientConfigurationNetworkHandler.method_57043(ClientConfigurationNetworkHandler.java:127)
	at knot//net.minecraft.client.network.ClientConfigurationNetworkHandler.openClientDataPack(ClientConfigurationNetworkHandler.java:115)
	at knot//net.minecraft.client.network.ClientConfigurationNetworkHandler.onReady(ClientConfigurationNetworkHandler.java:127)
	at knot//net.minecraft.network.packet.s2c.config.ReadyS2CPacket.apply(ReadyS2CPacket.java:22)
	at knot//net.minecraft.network.packet.s2c.config.ReadyS2CPacket.apply(ReadyS2CPacket.java:8)
	at knot//net.minecraft.network.NetworkThreadUtils.method_11072(NetworkThreadUtils.java:27)
	at knot//net.minecraft.util.thread.ThreadExecutor.executeTask(ThreadExecutor.java:164)
	at knot//net.minecraft.util.thread.ReentrantThreadExecutor.executeTask(ReentrantThreadExecutor.java:23)
	at knot//net.minecraft.util.thread.ThreadExecutor.runTask(ThreadExecutor.java:138)
	at knot//net.minecraft.util.thread.ThreadExecutor.runTasks(ThreadExecutor.java:123)
	at knot//net.minecraft.client.MinecraftClient.render(MinecraftClient.java:1318)
	at knot//net.minecraft.client.MinecraftClient.run(MinecraftClient.java:947)
	at knot//net.minecraft.client.main.Main.main(Main.java:265)
	at net.fabricmc.loader.impl.game.minecraft.MinecraftGameProvider.launch(MinecraftGameProvider.java:506)
	at net.fabricmc.loader.impl.launch.knot.Knot.launch(Knot.java:72)
	at net.fabricmc.loader.impl.launch.knot.KnotClient.main(KnotClient.java:23)
	at org.prismlauncher.launcher.impl.StandardLauncher.launch(StandardLauncher.java:102)
	at org.prismlauncher.EntryPoint.listen(EntryPoint.java:129)
	at org.prismlauncher.EntryPoint.main(EntryPoint.java:70)
```
The network protocol error is present in versions 1.21 and newer.

The problem accrues when the server sends an invalid wolf_variant registry data packet that contains entries with invalid tags, the vanilla client for 1.20.6 simply uses default values for those tags, but this behavior changed in 1.21, where instead the client ignores entries with invalid tags, and crashes if there's no entry with valid tags in them, creating a bug which can be exploited to allow vanilla 1.20.6 clients to join a server while not allowing newer clients using ViaProxy/ViaFabric/ViaFabricPlus/ViaForge to join.

To fix this, if the received registry data packet doesn't contain any valid entry, we construct and send a default registry data packet for wolf_variant registries, containing default values extracted from the notchain server, however if the received registry data packet contains at least one valid entry this patch does nothing.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_5to1_21/Protocol1_20_5To1_21.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_5to1_21/Protocol1_20_5To1_21.java
@@ -237,6 +237,7 @@ public final class Protocol1_20_5To1_21 extends AbstractProtocol<ClientboundPack
         addEntityTracker(connection, new EntityTrackerBase(connection, EntityTypes1_20_5.PLAYER));
         connection.put(new EfficiencyAttributeStorage());
         connection.put(new PlayerPositionStorage());
+        EntityPacketRewriter1_21.receivedValidWolfVariantRegistryData = false;
     }
 
     @Override
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_5to1_21/rewriter/EntityPacketRewriter1_21.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_5to1_21/rewriter/EntityPacketRewriter1_21.java
@@ -40,9 +40,11 @@ import com.viaversion.viaversion.protocols.v1_20_5to1_21.storage.PlayerPositionS
 import com.viaversion.viaversion.rewriter.EntityRewriter;
 import com.viaversion.viaversion.rewriter.RegistryDataRewriter;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 public final class EntityPacketRewriter1_21 extends EntityRewriter<ClientboundPacket1_20_5, Protocol1_20_5To1_21> {
-    public boolean receivedValidWolfArmorRegistryData = false;
+    public boolean receivedValidWolfVariantRegistryData = false;
 
     public EntityPacketRewriter1_21(final Protocol1_20_5To1_21 protocol) {
         super(protocol);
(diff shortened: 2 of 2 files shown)


kennytv (changes requested):
This doesn't look right, I don't see the behavior change you described. My guess is it's not sending registry data at all. At that point, you wouldn't have to send the full Vanilla registry either. plus, you can't use static variables here, rewriters can be used on multiple threads for multiple clients

amiralimollaei:
I think you're right, what happened was I accidentally mixed my log files with one from 1.21.8 and saw the *"invalid"* packet, but on 1.20.6 it doesn't send it at all, so all that needs to be done is to send the registry data packet for wolf_variant if its not sent by the server to fix the bug, i could modify my code, however the current patch also fixes the bug, just not efficiently, I will update the code as soon as possible.

amiralimollaei:
It seems like the fix was way simpler than I thought, the real problem was the wolf_variant RegistryData packet not being sent, while this is not required in 1.20.6 but it becomes mandatory for the server to send the wolf_variant RegistryData packet for newer versions and that's why the Network Protocol error happens when rewriting the packets from 1.20.6 to 1.21 or newer

amiralimollaei:
Sorry, I'm working on removing the static variable still

amiralimollaei:
I fixed the static variable, is there any other issue?

kennytv:
See the "storage" classes on how to store information for a particular connection, you can't put it into the rewriter since it is shared across connections. You also don't need to add all those wolf variants, a single one (pale) or possibly even none would be enough

amiralimollaei:
Hello, I've now implemented `WolfVariantRegistryCheckStorage` and also tried with sending no variant but it didn't work, so now it only sends one (pale) wolf variant.

amiralimollaei:
However, the fact that it still crashes with no variants sent, potentially can cause the same crash when the "wolf_variant" packet is sent by the server but it has no entries, but I think that might be intended behavior.

amiralimollaei:
Sorry, my IDE was freaking out about it, I removed the null checks, hopefully it's correct

PR #4659, opened 2025-10-02
refactor: improve some collection usages

MukjepScarlet:
1. No more LinkedList
2. Typed map and ref map (for Class keys)


MukjepScarlet:
Okay I'll change to `Int2xxxMap`s.

PR #4662, opened 2025-10-02
refactor: migrate HTTP requests to Java 11 HttpClient

MukjepScarlet:
New public API:
- GsonUtil.getPrettyGson -> a Gson instance with pretty printing
- interface ViaPlatform now extends Executor, to make it work with CompletableFuture. `execute` implementation is delegated to `runAsync`.
- HttpClientUtil provides some shortcut for JSON response handling and HttpClient constructing.

Breaking changes:
- GsonUtil now has no public constructor

Note: Because this change has no corresponding unit test, I have not verified it.

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/platform/ViaPlatform.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/platform/ViaPlatform.java
@@ -29,6 +29,7 @@ import com.viaversion.viaversion.api.connection.UserConnection;
 import java.io.File;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.concurrent.Executor;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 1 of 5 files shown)


florianreuth:
VV is downgraded to Java 8 builds on our CI by https://github.com/unimined/JvmDowngrader which afaik doesn't handle the Java 11 http client.

MukjepScarlet:
Okay, this will be removed then. I'd like to ask if other modifications useful?

florianreuth:
I don't know what Kenny thinks about this / I haven't had time to go through your PRs yet, but I would generally only make changes that actually improve performance or speed up loading. Going through old code and simply replacing APIs with “more modern” calls doesn't really offer any advantages imo.

MukjepScarlet:
Okay. I want to make interface `ViaPlatform` extends `Executor` (inspired by `MinecraftClient` implements `Executor` (render thread)), this makes it easier to work with CompletableFutures rather than nested `runAsync`.

florianreuth:
The scheduler functions in ViaPlatform are only used in the internals of VV/VB for mapping loading afaik, but I think there is no difference between calling runAsync() or execute() besides runAsync() being more informative on what the function actually does.

PR #4671, opened 2025-10-04
Rewrite item ids in 1.12.2->1.13 chat items

florianreuth:


kennytv (changes requested):
1.13 items can already be found in the identifiers-1.13 file / from the 1.13->1.13.2 protocol

PR #4672, opened 2025-10-04
Don't re-use the 1.13 component rewriter for the 1.14 protocol

FlorianMichael:


kennytv:
This should go into the other PR as it is unnecessary on its own, but is required for 4671

PR #4673, opened 2025-10-04
Fix "Filter Invalid Teams Packet" fix storing teams of players which the client does not store

LeonTG:
In the vanilla client, if it receives a "Add Player to Team" packet, it will automatically remove the player from any other team the client thinks the player is on

However, the fix from PR #4643 does not do this, and is storing the team of the player when the player is possibly not on that team by the clients knowledge

So I figured I would make a PR to address this minor thing


FlorianMichael (approved):
Oh I oversaw that, thank you!

PR #4700, opened 2025-11-01
Handle early connection closure for bot attacks

linsaftw:
Add handling for early connection closure in bot attacks.

This is key and essential to support [VeloFlame](https://builtbybit.com/resources/veloflame-secure-minecraft-proxy.80990/), it will block the bot early and not inject it, meaning, packet decoders will not be available for ViaVersion, and throw an exception. This patch is key because it prevents viaversion from triggering on velocity bot attacks, when VeloFlame is installed it makes sure memory is lean and not inject useless handlers when already closed.


kennytv:
You should be clearing the pipeline yourself if you kill it at an unexpected time, or skip calling init yourself

linsaftw:
> You should be clearing the pipeline yourself if you kill it at an unexpected time, or skip calling init yourself

What?

The pipeline is clear and empty, viaversion is trying to inject to a empty pipeline, and spits errors!

Please apply the merge to solve. Thank you.

linsaftw:
I made a mistake. Please check again!

PR #4701, opened 2025-11-01
When original initializer closes the connection, immediately stop

linsaftw:
This correctly adddresses the issue of the original initializer closing the connection, it will now immediately stop properly unlike my other faulty PR. Thank you for pointing out!


linsaftw:
image
Working! No errors now!

florianreuth:
You should still wrap the current initializers into your own object to prevent the initChannel() call. See ConnectionManager#serverChannelInitializer and ConnectionManager#backendChannelInitializer.

linsaftw:
> You should still wrap the current initializers into your own object to prevent the initChannel() call. See ConnectionManager#serverChannelInitializer and ConnectionManager#backendChannelInitializer.

Thank you so much. We instead implemented your workaround. 100% working!

PR #4716, opened 2025-11-21
Fix 1.21.9->1.21.7 entity id mapping

Lumine1909:
Previously, entity id mapping is incorrect (such as a 1.21.10 client side `creeper` will be map to a 1.21.8 server side `dark_oak_boat`) since mojang's update in 1.21.9.

To test, you can get a item with this nbt in creative mode `{components:{"minecraft:entity_data":{id:"creeper"}},count:1,id:"minecraft:bat_spawn_egg"}`

Command: `/give @a minecraft:bat_spawn_egg[minecraft:entity_data={id:"creeper"}]`

And then press E twice to open and close the inventory (let client send a `ServerboundSetCreativeModeSlotPacket`), then the spawn egg will spawn a `dark_oak_boat` instead.

Video: https://github.com/user-attachments/assets/e58964d8-f0bb-43d7-8f45-d0ef8dc9e143

This pr fix that mapping bug.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_21_7to1_21_9/rewriter/BlockItemPacketRewriter1_21_9.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_21_7to1_21_9/rewriter/BlockItemPacketRewriter1_21_9.java
@@ -18,7 +18,6 @@
 package com.viaversion.viaversion.protocols.v1_21_7to1_21_9.rewriter;
 
 import com.viaversion.viaversion.api.connection.UserConnection;
-import com.viaversion.viaversion.api.data.FullMappings;
 import com.viaversion.viaversion.api.data.entity.EntityTracker;
 import com.viaversion.viaversion.api.minecraft.ResolvableProfile;
 import com.viaversion.viaversion.api.minecraft.data.StructuredDataContainer;


Lumine1909:
> You only have to swap `mappedIdentifier` with `identifier` in the original code, as it's called after mapping the ids and that's the fast way of going from int id to string

No that won't work.
From the current mapping system, the `identifier` and `mappedIdentifier` is look like that:
```
30, idStr: minecraft:creeper, mappedIdStr: minecraft:creaking
31, idStr: minecraft:dark_oak_boat, mappedIdStr: minecraft:creeper
32, idStr: minecraft:dark_oak_chest_boat, mappedIdStr: minecraft:dark_oak_boat
33, idStr: minecraft:dolphin, mappedIdStr: minecraft:dark_oak_chest_boat
```
When `mappedIdStr` is `dark_oak_boat`, the original `id` is 32, and the the `idStr` is `dark_oak_chest_boat`.
But what we need here is `creeper`, which is `idStr` of 30 and `mappedIdStr` of 31, we need `inverse` it back (not the `inverse` method, it also didn't work), my code is the most concise way that I can find without modify mapping system.

And what swap `mappedIdentifier` with `identifier` do is exactly map client `creeper` to server `dark_oak_chest_boat`

Lumine1909:
> > You only have to swap `mappedIdentifier` with `identifier` in the original code, as it's called after mapping the ids and that's the fast way of going from int id to string
> 
> No that won't work. From the current mapping system, the `identifier` and `mappedIdentifier` is look like that:
> 
> ```
> 30, idStr: minecraft:creeper, mappedIdStr: minecraft:creaking
> 31, idStr: minecraft:dark_oak_boat, mappedIdStr: minecraft:creeper
> 32, idStr: minecraft:dark_oak_chest_boat, mappedIdStr: minecraft:dark_oak_boat
> 33, idStr: minecraft:dolphin, mappedIdStr: minecraft:dark_oak_chest_boat
> ```
> 
> When `mappedIdStr` is `dark_oak_boat`, the original `id` is 32, and the the `idStr` is `dark_oak_chest_boat`. But what we need here is `creeper`, which is `idStr` of 30 and `mappedIdStr` of 31, we need `inverse` it back (not the `inverse` method, it also didn't work), my code is the most concise way that I can find without modify mapping system.
> 
> And what swap `mappedIdentifier` with `identifier` do is exactly map client `creeper` to server `dark_oak_chest_boat`

Btw the 30-33 is generated by
```
        var mapping = Protocol1_21_7To1_21_9.MAPPINGS.getEntityMappings();
        for (int i = 0; i < mapping.mappedSize(); i++) {
            System.out.println(i + ", idStr: " + mapping.identifier(i) + ", mappedIdStr: " + mapping.mappedIdentifier(i));
        }
```

kennytv:
Ah that's because there is another issue in `EntityData`. If you fix the first line in its `rewrite` method to be `Rewritable.entityRewriteFunction(protocol, clientbound).applyAsInt(type)`, then you can change the block item rewriter to be what I suggestes in my review

Lumine1909:
> Rewritable.entityRewriteFunction(protocol, clientbound).applyAsInt(type)

Great it works!

PR #4725, opened 2025-11-29
Mitigate security vulnerabilities in CI

Kichura:
1. Adds a [cooldown system](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#cooldown-) for Dependabot to freeze version updates, *(exception being security updates)*
2. Changes the CI's contents permission to read-only - However use `write` if a specific repository absolutely [requires it](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents), *(like in update-gradle-wrapper workflow)*
3. Pins all job versions to their commit hashes to avoid [supply chain attacks towards the workflow script](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash).

*(All security issues were spotted by the `zizmor` library)*

Also. not part of security but `fetch-depth: 0` is restored as `update-gradle-wrapper` workflow depends on it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,4 +5,4 @@ updates:
       default-days: 7
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"


.github/workflows/build.yml
```diff
@@ -1,23 +1,22 @@
-name: Java CI with Gradle
-
-on: [push, pull_request, workflow_dispatch]
+name: Build
+on: [pull_request, push, workflow_dispatch]
+permissions:
+  contents: read
```
kennytv: The Upload Artifacts to GitHub step needs more permissions

.github/dependabot.yml
```diff
@@ -1,6 +1,8 @@
 version: 2
 updates:
   - package-ecosystem: "github-actions"
+    cooldown:
+      default-days: 7
     directory: "/"
     schedule:
       interval: "weekly"
```
kennytv: Change this to monthly to reduce the update spam that comes from specific subversions

PR #4737, opened 2025-12-13
chore: make 1.8.x blocking component clearer

MukjepScarlet:
Meanwhile prevent to create too many same objects.


florianreuth:
Why? this shouldn't be a field in the API and re-creating it isn't that bad

kennytv:
It's technically mutable and doesn't do all too much, so I'm closing this

PR #4782, opened 2026-01-11
Workaround Fabric Particle API modifying vanilla packets

florianreuth:
This patch disables Fabrics Particle API as they break vanilla packet structure which causes exceptions in our rewriter handlers.


Kichura (approved):

florianreuth:
going to move the code into the velocity submodule soon:tm:

florianreuth:
Requires https://github.com/PaperMC/Velocity/pull/1706 to be merged

PR #4796, opened 2026-01-18
Handle single item component -> multi item component empty translation

RaphiMC:
The 1.21 food component got replaced by 3 components. This PR adds handling for when the food component was empty and sets the other 3 also to empty

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/minecraft/data/StructuredDataContainer.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/minecraft/data/StructuredDataContainer.java
@@ -177,11 +177,10 @@ public final class StructuredDataContainer implements Copyable {
                 set(toKey, replacement);
             }
         } else {
+            // Also replace the key for empty data
+            setEmpty(toKey);
             if (emptyValueHandler != null) {
                 emptyValueHandler.run();
-            } else {
-                // Also replace the key for empty data
-                setEmpty(toKey);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


FlorianMichael (approved):

api/src/main/java/com/viaversion/viaversion/api/minecraft/data/StructuredDataContainer.java
```diff
@@ -173,8 +177,12 @@ public <T, V> void replace(final StructuredDataKey<T> key, final StructuredDataK
                 set(toKey, replacement);
             }
         } else {
-            // Also replace the key for empty data
-            setEmpty(toKey);
+            if (emptyValueHandler != null) {
+                emptyValueHandler.run();
+            } else {
+                // Also replace the key for empty data
+                setEmpty(toKey);
```
kennytv: I think this should always be called

kennytv: Then "replace" and passing an unused replacement doesn't really make sense

kennytv:
If you still want your initial idea in, you can make the runnable a supplier that still allows the result to be empty, removed, or an actual value, like
```java
public interface EmptyMapping<V> {
    static <V> EmptyMapping<V> empty() { return new EmptyMappingImpl<>(Result.SET_EMPTY, null); }
    static <V> EmptyMapping<V> remove() { return new EmptyMappingImpl<>(Result.REMOVE, null); }
    static <V> EmptyMapping<V> value(V value) { return new EmptyMappingImpl<>(Result.SET_VALUE, value); }
}
```
or have the impl hold a nullable optional instead of the Result type, but that's more cursed

RaphiMC:
Its technically still possible to do, what I originally planned for, because setEmpty is called before the Runnable is called, so the Runnable could just override it if needed

kennytv (approved):
Yeah good enough, that would be more niche anyway

PR #4813, opened 2026-01-29
Add null checks for packetValues and readableObjects

WhiteProject1:
Updated PacketWrapperImpl to handle cases where packetValues and readableObjects may be null, preventing potential NullPointerExceptions. This includes initializing these collections when needed and adding appropriate null checks throughout the class.


kennytv:
> Updated PacketWrapperImpl to handle cases where packetValues and readableObjects may be null, preventing potential NullPointerExceptions.

Why? They weren't nullable until you made them with this PR

WhiteProject1:
I just change for GC

kennytv:
The memory impact here is negligible, if anything this can make predictive optimization harder

PR #4820, opened 2026-01-30
Fix play state compression in 1.8->1.9

florianreuth:
Fixes play state compression enabling when the login state compression was already enabled, causing the compression handlers to be unchecked cast against Via's handler classes.


kennytv (changes requested):

kennytv:
Only removing them if they aren't our own compressors if >= 0 and otherwise using the threshold setter is better than always replacing

kennytv (approved):

PR #4821, opened 2026-01-31
Allow platforms to change compression names

florianreuth:


api/src/main/java/com/viaversion/viaversion/api/platform/ViaInjector.java
```diff
@@ -79,7 +79,7 @@ default SortedSet<ProtocolVersion> getServerProtocolVersions() throws Exception
     }
 
     /**
-     * Get the name of the encoder for then netty pipeline for this platform.
```
Kichura: why was this changed from "the netty pipeline" to "then netty pipeline" ?

PR #4869, opened 2026-03-29
Fix NPE in container click of item with the 'blocks_attacks' attribute

Galapen:
Handles the possibly null `tagKey` of optional `bypassed_by` in `BlocksAttacks`.

Fixes #4868


kennytv:
Thanks, but it was already fixed in VB - this solution wouldn't be right, the value should never be null

PR #4894, opened 2026-04-02
Fix NPE in SharedRegistrations when protocol versions are null

ohnodev:
## Summary
- add null guards for `protocol.getServerVersion()` and `protocol.getClientVersion()` in `SharedRegistrations.applyMatching()` to prevent startup/runtime NPEs when platform wrappers provide null versions
- keep behavior unchanged for valid versions while still bailing out for mismatched version types
- add debug-gated warning logging for null-version bail-outs to make integration diagnostics easier without adding noise in normal operation

## Test plan
- [x] Reproduced NPE path with ViaFabric + ViaVersion `5.9.0-SNAPSHOT` where protocol versions can be null
- [x] Verified server no longer crashes in `SharedRegistrations.applyMatching`
- [x] Verified affected change scope is one file (`SharedRegistrations.java`) with minimal diff

Made with [Cursor](https://cursor.com)

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocol/shared_registration/SharedRegistrations.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocol/shared_registration/SharedRegistrations.java
@@ -17,6 +17,7 @@
  */
 package com.viaversion.viaversion.protocol.shared_registration;
 
+import com.viaversion.viaversion.api.Via;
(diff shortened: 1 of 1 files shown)


florianreuth:
ViaFabric and ViaFabricPlus will be updated properly to the dev branch so that case doesn't appear in the first place.

PR #4910, opened 2026-04-15
fix block state translation in falling block

### LINKED ISSUE #4908: On a 1.20.6 server, when using a client version 1.21.2 or above, falling blocks display incorrectly.
### `/viaversion dump` Output

https://dump.viaversion.com/1ec5240150b3e9275acae489df6ea3a295d311e1f4e33ab44647dd2a217f1713

### Server/Client Error

none

### Bug Description

On a 1.20.6 server, when logging in with a client version 1.21.2 or above and placing falling blocks (such as sand, gravel, etc.), issues may occur where entities disappear or display incorrectly.
1.20.6
https://github.com/user-attachments/assets/310e555d-3611-4297-8c70-6a9df902e784
1.21.2+
https://github.com/user-attachments/assets/827c79c2-2af0-4053-b640-5afeb9d2a44f

### Steps to Reproduce

1. Set up a 1.20.6 server.
2. Log in to the server using clients from versions 1.20.6–1.21.1, and 1.21.2 or above.
3. Place falling blocks and observe the falling process.

### Additional Server Info

_No response_

### Checklist

acrylic-style:
Fixes #4908

The code might not be very efficient, but it works fine.

Tested on server 1.20.6 and client 1.21.11 + 26.1.2:
https://dump.viaversion.com/de0e6025cef1186c5473814ccf094cc8ba9100684caaf0fa304177dd5c04dcfd

https://github.com/user-attachments/assets/f7381d74-1680-401e-8894-2ba2c3a3b677

I also have tested on server 1.15.2 and client 1.21.11 and it worked fine.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_16_4to1_17/rewriter/EntityPacketRewriter1_17.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_16_4to1_17/rewriter/EntityPacketRewriter1_17.java
@@ -69,6 +69,15 @@ public final class EntityPacketRewriter1_17 extends EntityRewriter<ClientboundPa
             wrapper.cancel();
             setDirection.send(Protocol1_16_4To1_17.class);
         });
+        protocol.appendClientbound(ClientboundPackets1_16_2.ADD_ENTITY, wrapper -> {
+            final int entityType = wrapper.get(Types.VAR_INT, 1);
+            if (entityType != EntityTypes1_17.FALLING_BLOCK.getId()) {
+                return;
+            }
+            final int data = wrapper.get(Types.INT, 0);
+            final int mappedBlockStateId = protocol.getMappingData().getNewBlockStateId(data);
+            wrapper.set(Types.INT, 0, mappedBlockStateId);
+        });
 
         protocol.registerClientbound(ClientboundPackets1_16_2.REMOVE_ENTITIES, null, wrapper -> {
             int[] entityIds = wrapper.read(Types.VAR_INT_ARRAY_PRIMITIVE);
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/v1_16_4to1_17/rewriter/EntityPacketRewriter1_17.java
```diff
@@ -69,6 +69,15 @@ public void registerPackets() {
             wrapper.cancel();
             setDirection.send(Protocol1_16_4To1_17.class);
         });
+        protocol.appendClientbound(ClientboundPackets1_16_2.ADD_ENTITY, wrapper -> {
```
florianreuth: Pretty sure this is not needed as already done via shared registrations.

common/src/main/java/com/viaversion/viaversion/rewriter/EntityRewriter.java
florianreuth, line 318: What about this function? Probably the reason the code in 1.16.4->1.17 was needed

florianreuth (changes requested):
Not sure what @kennytv says here, but I would prefer a mappedTypeFromId in EntityRewriter.

common/src/main/java/com/viaversion/viaversion/rewriter/EntityRewriter.java
acrylic-style: I missed that function because I was sleepy (3 am here), but now this is fixed

florianreuth:
Hm I think you can just change id(type); to mappedId(type); in EntityRewriter#typeFromId and get rid of the new function

florianreuth:
I think we should just keep the typeFromId name, as the typeFromId(int type) also doesn't use the mapped term

florianreuth (approved):

kennytv (changes requested):
Keep the old name, move the impl to EntityRewriter to also check for its `trackMappedType` field (`trackMappedType ? mappings.mappedId(type) : mappings.id(type)`)

kennytv (approved):
Thanks!

PR #4917, opened 2026-04-22
Fix version comments, Unsubscribe from proprietary caching tier

Kichura:
1. Fixes version comments by including `v` at beginning of them to resolve zizmor's versioning mismatch.
2. Unsubscribes from Gradle's enhanced caching provider as it is proprietary and requires agreeing to Gradle's TOS.


kennytv:
Caching shouldn't be disabled

Kichura:
> Caching shouldn't be disabled

Did this intentionally as caching has a chance of causing builds to fail due to update conflicts. but can re-enable if it isn't an issue for this repo.

kennytv (approved):

PR #4923, opened 2026-04-26
Migrate chunk-key maps to fast util

Pablete1234:
I found these mostly by accident when investigating a heap dump for an unrelated issue, but java `HashMap`s get terrible hashing for these `long` based chunk keys, causing most of the nodes in the map not to be `o(1)` direct nodes, but instead be either linear searches or binary maps for keys. This is the result of how the keys are constructed and "mixed".

The chunk keys are one long with 32 bits for chunkX, and 32 bits for chunkZ. when these get hashed, java's default hashCode for longs is `value ^ (value >>> 32)`, which effectively xors x and z directions, this is further aggravated by the "mixing" of the key done in hashmap, which then does `hash ^ (hash >>> 16)`. Given *most* chunks are mostly in the small number rage (eg: -128 to +128 range), this repeated shifting of values ends up with very, very little usable keys.

### Verifying/Testing:
I've created a test where i hashed all chunks between -128 and +128 for x and z directions, in a table size of 1024 (note the table size of 1024 chunks is reasonable, every single chunk between -128 and +128 isn't, but the point is to see how they'd spread across the space).
Java's HashMap only uses 32 buckets with 512 each, leaving empty the other ~1000 buckets. FastUtil's impl's top 32 buckets hold 88 to 77 (best case would be 16 per bucket)
Source and result can be found at https://gist.github.com/Pablete1234/47fc7561ece621aca7b054b80d77264c

### DIFF
--- a/build-logic/src/main/kotlin/via.shadow-conventions.gradle.kts
+++ b/build-logic/src/main/kotlin/via.shadow-conventions.gradle.kts
@@ -49,7 +49,7 @@ fun ShadowJar.configureRelocations() {
 }
 
 fun ShadowJar.configureExcludes() {
-    // FastUtil - we only want object, int, and certain reference maps
+    // FastUtil - we only want object, int, long-to-object, and certain reference maps
     // Object types
     exclude("it/unimi/dsi/fastutil/*/*2Reference*")
     exclude("it/unimi/dsi/fastutil/*/*Reference2Int*")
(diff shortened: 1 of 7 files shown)


kennytv:
I don't think this is currently shaded, so it'd only be available on platforms that have the full library

Pablete1234:
A bunch of things in the codebase are at the very least using imports for it:
https://github.com/search?q=repo%3AViaVersion%2FViaVersion+fastutil&type=code

it seems that its shaded, but in a stripped down version:
https://github.com/ViaVersion/ViaVersion/blob/567ef33f6f5a295baa269ae907a32e0d7e6ea492/build-logic/src/main/kotlin/via.shadow-conventions.gradle.kts#L61

would you rather i make it only load on platforms that support it, or stop removing the `long` package of the lib?

kennytv:
Here's exclusions that make the increase reasonably small
```patch
 fun ShadowJar.configureExcludes() {
-    // FastUtil - we only want object, int, and certain reference maps
+    // FastUtil - we only want object, int, long-to-object, and certain reference maps
     // Object types
     exclude("it/unimi/dsi/fastutil/*/*2Reference*")
     exclude("it/unimi/dsi/fastutil/*/*Reference2Int*")
@@ -58,13 +58,17 @@ fun ShadowJar.configureExcludes() {
     exclude("it/unimi/dsi/fastutil/*/*Short*")
     exclude("it/unimi/dsi/fastutil/*/*Float*")
     exclude("it/unimi/dsi/fastutil/*/*Double*")
-    exclude("it/unimi/dsi/fastutil/*/*Long*")
+    exclude("it/unimi/dsi/fastutil/*/*2Long*")
     exclude("it/unimi/dsi/fastutil/*/*Char*")
     // Map types
     exclude("it/unimi/dsi/fastutil/*/*Custom*")
     exclude("it/unimi/dsi/fastutil/*/*Tree*")
     exclude("it/unimi/dsi/fastutil/*/*Heap*")
     exclude("it/unimi/dsi/fastutil/*/*Queue*")
+    exclude("it/unimi/dsi/fastutil/longs/*2Int*")
+    exclude("it/unimi/dsi/fastutil/longs/*ArrayMap*")
+    exclude("it/unimi/dsi/fastutil/longs/*LinkedOpenHashMap*")
+    exclude("it/unimi/dsi/fastutil/longs/*SortedMap*")
     // Crossing fingers
     exclude("it/unimi/dsi/fastutil/*/*Big*")
     exclude("it/unimi/dsi/fastutil/*/*Synchronized*")
```
then you can remove the BlockConnectionStorage reflection and also make uses of the long maps/sets the proper fastutil interfaces. make sure to check getters and iterators too to use the non-boxing methods

kennytv (approved):

PR #4936, opened 2026-05-08
Include stacktrace in gradlew build

Kichura:
Title says it all.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
         java-version: 17
         check-latest: true
     - name: Build with Gradle
-      run: ./gradlew build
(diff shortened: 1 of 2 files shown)


florianreuth:
Not sure if this is really needed, usually when building fails you end up doing local testing with --stacktrace and not on the actual GH repo.

Kichura:
> Not sure if this is really needed, usually when building fails you end up doing local testing with --stacktrace and not on the actual GH repo.

I included it anyways in case the build suddenly fails in GH actions and you need to stacktrace it without having to do it locally, otherwise i would've never done it for baseproject.

Kichura:
Closing as stale.

PR #4946, opened 2026-05-24
[Fix] Crash realated to missing registry keys/entries

amiralimollaei:
The `sendMissingRegistries` only sends the missing registry entries that have been missing in **all previews user connections**.

**Steps To Reproduce on a 26.1 Client:**

1.  join a 1.21.11 server that sends all the registries (Vanilla)
2.  then join a 1.21.11 server that has missing registries (LimboAPI)

The function `sendMissingRegistries` in `RegistryDataRewriter`, checks if a registry is already sent before "FinishConfiguration" by using `registryKeyMappings.containsKey(entry.getKey())`, However `registryKeyMappings` is shared between all connections, because `RegistryDataRewriter` is only instantiated once, this causes a crash with missing registry.

To fix this, we now use a `StorableObject` and for every connection that uses the handle we record this value separately using `wrapper.user().put(new RegistryKeyMapping())`.

To also keep the old behavior for parts of the code that relied on it, we also keep the same shared `registryKeyMappings` as well, but use our `RegistryKeyMapping` object only in `sendMissingRegistries`.

I also added the missing `temperate` chicken variant to the `chicken_variant` registry.


Kichura:
Something i wanna point out, This is not a server crash as otherwise this would've caused all server players to be kicked out at once.

amiralimollaei:
yes, this only happens on the client

florianreuth:
Didn't check yet if this is correct at all, but is it maybe better to have them stored inside the EntityTrackerBase?

amiralimollaei:
I tried to make a minimal PR, moving it to `EntityTrackerBase` (even though I fail to realize why registry entries should be tracked inside the entity tracker) would require much more refactoring, and since I'm not familiar with the rest of the codebase or your coding styles, I doubt it would be acceptable.

kennytv (approved):
I moved it around

PR #4952, opened 2026-05-31
Fix legacy skull rotation overflow in 1.12.2 to 1.13 translation

itslilscorp:
## Description

Clamps legacy standing skull rotations to the valid `0..15` range when translating 1.12.2 skull block entities to 1.13 block states.

Also avoids applying the `Rot` tag to wall-mounted skulls, since wall skulls use facing states rather than rotation states.

## Problem

Some custom player heads render correctly on a 1.12.2 client but appear as creeper or dragon heads on clients using 1.13 or newer protocols.

In 1.13, skull variants occupy adjacent block-state ID ranges. The previous translation added the legacy `Rot` value directly to the mapped state ID.

Out-of-range values such as `20` could therefore spill into the next skull variant range.

For example:

`player_head + 20 -> creeper_head`

## Screenshots

1.12.2 client connected to the 1.12.2 server:
image

1.13+ client connected to the same server before this patch:

1.13+ client connected to the same server after this patch:

## Testing

Added regression tests covering:

- An out-of-range standing player-head rotation wrapping into `0..15`
- A wall-mounted player head ignoring `Rot`

Verified with:

`./gradlew :viaversion-common:test`

#### AI was used to help investigate the issue and prepare the patch. The behavior was reproduced and the resulting change was manually tested.


common/src/main/java/com/viaversion/viaversion/protocols/v1_12_2to1_13/provider/blockentities/SkullHandler.java
```diff
@@ -41,15 +42,15 @@ public int transform(UserConnection user, CompoundTag tag) {
 
         int id = storage.get(position).getOriginal();
         if (id >= SKULL_WALL_START && id <= SKULL_END) {
+            NumberTag rot = tag.getNumberTag("Rot");
+            if (id == SKULL_FLOOR_START && rot != null) {
```
kennytv: Remove the skull foor start number, that looks both incorrect and not really unnecessary

kennytv (changes requested):
You can remove the test class too, it's good for verification but doesn't need to be there post merge

itslilscorp:
There you go! I removed both and confirmed everything seems to be as expected in my test area.

kennytv (approved):

PR #4961, opened 2026-06-05
fix https://github.com/ViaVersion/ViaVersion/issues/4959

trassert:
https://github.com/ViaVersion/ViaBackwards/issues/1292#issuecomment-4632714075


kennytv:
the nether does not have a clock

ShawezA:
The fix is needed because in 26.1, dimension_type entries are expected to have a direct clock field alongside timelines. When translating from 1.21.11, that field is absent for nether/end since 1.21.11 didn't have it. The 26.1 client falls back to using the dimension's resource key as the clock lookup when the field is missing, which means it looks for minecraft:the_nether and minecraft:the_end in the world_clock registry. Without those entries the client fails registry validation on connection.

The empty CompoundTag entries act as zero-tick-rate clocks, which is functionally equivalent to has_fixed_time, the nether doesn't tick, but the registry slot needs to exist for the client to proceed. 

An alternative would be to explicitly add clock = "overworld" to the nether/end dimension_type entries in the handler, which would remove the need for these world_clock entries entirely.

kennytv:
> The 26.1 client falls back to using the dimension's resource key as the clock lookup

I don't think that's true, where did you get this from/see this? The PR was ok aside from the fact the nether clock is unnecessary. It does need the end clock though

ShawezA:
Sorry about that I just guessed lol. I tested it with just the end clock and its working as well, so it was probably that

kennytv (approved):
Thanks

kennytv:
Doesn't look related to the linked issue, just your comment (still being a valid issue, just a different one), so I'll leave that open

PR #4964, opened 2026-06-07
Fix legacy nausea effect intensity on 1.21.5+ clients

itslilscorp:
### What changed
Sets the mob effect blend flag for nausea when translating `UPDATE_MOB_EFFECT` from 1.21.4 to 1.21.5.

### Why
Nausea has explicit blend timing in 1.21.5. Without the blend flag, modern clients skip the transition and short/low-intensity nausea effects can appear at full strength immediately.

This preserves all existing mob effect flags and only adds the blend flag for nausea.

### Testing
- Verified on a 1.12.2 server with a modern client that short nausea effects no longer render at full strength immediately.

#### AI was used to help investigate the issue and prepare the patch. The behavior was reproduced and the resulting change was manually tested.


kennytv (changes requested):
This doesn't work, the id is wrong. You need to actually verify AI output

itslilscorp:
Ah, that's my bad. Before narrowing the fix down to just changing nausea, the AI previously incorrectly applied this to all effects. So, it looks like when I was verifying this build, I left that old version in my testing environment by accident, so as far as I knew, it "worked". I've updated the id to properly target nausea AND tested with the actual proper jar this time. :)

kennytv (approved):
Thanks

PR #4975, opened 2026-06-19
Add limit to debug sample subscription count

Pantera07:
A vanilla client's `subscription count` is expected to be 1. 
This PR adds a maximum limit of 64 to prevent massive `for` loops.


common/src/main/java/com/viaversion/viaversion/protocols/v1_21_7to1_21_9/Protocol1_21_7To1_21_9.java
```diff
@@ -112,7 +113,7 @@ protected void registerPackets() {
             wrapper.write(Types.VAR_INT, 0); // Number of block particles
         });
         registerServerbound(ServerboundPackets1_21_6.DEBUG_SAMPLE_SUBSCRIPTION, wrapper -> {
-            final int count = wrapper.read(Types.VAR_INT); // subscription count
+            final int count = Limit.max(wrapper.read(Types.VAR_INT), 64); // subscription count
```
kennytv: Should be 32 to mirror vanilla

Pantera07: Done. I've updated the subscription count limit to 32 to match vanilla.

kennytv (approved):

PR #4978, opened 2026-06-19
use java CRC32C impl to hash bytes instead of a custom one

OverwriteMC:
Motivation:

This approach is simply more optimized.
Based on the basic JMH test, this approach is approximately 40 times faster.

```
Benchmark          (size)  Mode  Cnt     Score     Error  Units
Main.testJava        1024  avgt    3    47,200 ±  21,505  ns/op
Main.testOriginal    1024  avgt    3  1791,135 ± 141,751  ns/op
```

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/codec/hash/CRC32C.java
+++ b/common/src/main/java/com/viaversion/viaversion/codec/hash/CRC32C.java
@@ -19,30 +19,10 @@ package com.viaversion.viaversion.codec.hash;
 
 final class CRC32C implements HashFunction {
 
-    private static final int[] CRC32C_TABLE = new int[256];
-
-    static {
-        for (int i = 0; i < 256; i++) {
-            int crc = i;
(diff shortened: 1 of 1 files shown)


florianreuth:
We can't use that as we downgrade ViaVersion to Java 8 using https://github.com/unimined/JvmDowngrader

OverwriteMC:
> We can't use that as we downgrade ViaVersion to Java 8 using https://github.com/unimined/JvmDowngrader

Interesting
It's strange that the tests don't take this into account
But then
Can we choose what will work depending on the Java version?
That would be nice solution

kennytv:
I'd prefer using the old solution when a simple class reflection check for CRC32C/Java version check fails. Unsafe use we have to maintain ourselves is a nono for something like this when most people, the ones we properly support, can use the direct solution

kennytv:
https://github.com/ViaVersion/ViaVersion/commit/d2a715862aab9a87618fa11954f88265e65ca535

PR #4984, opened 2026-06-26
Fix JsonParseException when downgrading malformed LOCK component in 1…

EsMokyGame:
….21.2

When a 1.21.2 client sends a `CONTAINER_CLICK` packet for an item (like a Shulker Box) that contains a `minecraft:lock` component with a null or empty name (often used by GUI plugins to lock containers), `customName.getValue()` can be the exact string `"null"`.

During the downgrade process in `BlockItemPacketRewriter1_21_2#downgradeItemData`, `SerializerVersion.V1_20_5.toComponent()` attempts to parse `"null"` as JSON, which Gson evaluates to `JsonNull`. This throws a `MergedCodecException` (`JsonParseException`) because the text component codec expects a String, Object, or Array. This exception drops the server connection with a `DecoderException`.

This commit adds a safeguard to explicitly check for `"null"` strings and wraps the component deserialization in a `try-catch` block to gracefully handle malformed JSON strings. It returns an empty lock tag instead of crashing the client connection.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/BlockItemPacketRewriter1_21_2.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/BlockItemPacketRewriter1_21_2.java
@@ -588,16 +588,11 @@ public final class BlockItemPacketRewriter1_21_2 extends StructuredItemRewriter<
 
             // Back from json in the string tag to plain text
             final StringTag customName = TagUtil.getNamespacedStringTag(itemComponentsTag, "custom_name");
-            if (customName == null || customName.getValue() == null || customName.getValue().equals("null")) {
+            if (customName == null || customName.getValue().isEmpty() || customName.getValue().equals("null")) {
                 return null;
             }
 
-            try {
-                return new StringTag(SerializerVersion.V1_20_5.toComponent(customName.getValue()).asUnformattedString());
-            } catch (Exception e) {
-                // If it fails to parse (e.g., malformed JSON), return empty lock to prevent client disconnect
-                return new StringTag("");
-            }
+            return new StringTag(SerializerVersion.V1_20_5.toComponent(customName.getValue()).asUnformattedString());
         });
         dataContainer.replace(StructuredDataKey.INSTRUMENT1_21_2, StructuredDataKey.INSTRUMENT1_20_5, instrument -> {
             if (instrument.hasId()) {
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/BlockItemPacketRewriter1_21_2.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/BlockItemPacketRewriter1_21_2.java
@@ -588,11 +588,16 @@ public final class BlockItemPacketRewriter1_21_2 extends StructuredItemRewriter<
 
             // Back from json in the string tag to plain text
             final StringTag customName = TagUtil.getNamespacedStringTag(itemComponentsTag, "custom_name");
-            if (customName == null) {
+            if (customName == null || customName.getValue() == null || customName.getValue().equals("null")) {
                 return null;
             }
 
-            return new StringTag(SerializerVersion.V1_20_5.toComponent(customName.getValue()).asUnformattedString());
+            try {
+                return new StringTag(SerializerVersion.V1_20_5.toComponent(customName.getValue()).asUnformattedString());
+            } catch (Exception e) {
+                // If it fails to parse (e.g., malformed JSON), return empty lock to prevent client disconnect
+                return new StringTag("");
+            }
         });
         dataContainer.replace(StructuredDataKey.INSTRUMENT1_21_2, StructuredDataKey.INSTRUMENT1_20_5, instrument -> {
             if (instrument.hasId()) {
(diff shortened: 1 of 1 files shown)


kennytv:
Does the client not disconnect on invalid strings (outside of null)? if not, you cannot add a try catch there. also, getValue can't return null there

EsMokyGame:
Thanks for the quick response!

Regarding `getValue() == null`, you are absolutely right, `StringTag.getValue()` doesn't return null, that check was redundant on my end.

Regarding disconnecting on invalid strings: The client itself doesn't disconnect when interacting with the item; instead, the vanilla 1.21.2 client sends the `CONTAINER_CLICK` packet back to the server containing the `minecraft:lock` component with its `custom_name` string value set exactly to `"null"`. 

This happens consistently on our 1.12.2 server with 1.21.2 clients when players click on a Shulker Box inside a GUI (which likely has `Lock: ""` or `Lock: "null"` assigned by the GUI plugin to prevent opening). 

Because the vanilla client is the one sending this `"null"` string back to the server, `SerializerVersion.V1_20_5.toComponent()` parses `"null"` as `JsonNull`, which causes an uncaught `MergedCodecException` during `downgradeItemData` and drops the connection for legitimate players.

Here is the exact crash log we are getting when a player clicks the Shulker Box GUI item:

<details>
<summary>Click to expand Stacktrace</summary>

<pre><code>
[01:05:29 WARN]: [ViaVersion] ERROR IN Protocol1_21To1_21_2 IN REMAP OF CONTAINER_CLICK (0x10)
[01:05:29 WARN]: io.netty.handler.codec.DecoderException: com.viaversion.viaversion.exception.InformativeException: Please report the full error (!) from the console on the Via support Discord or open an issue on the relevant GitHub repository.
[01:05:29 WARN]: Source 0: com.viaversion.viaversion.protocols.v1_21to1_21_2.rewriter.BlockItemPacketRewriter1_21_2$$Lambda/0x00000000737a2e48
...
[01:05:29 WARN]: Caused by: com.viaversion.viaversion.libs.gson.JsonParseException: com.viaversion.viaversion.libs.mcstructs.converter.model.MergedCodecException: Failed to deserialize as either left or right: [Failed to deserialize as either left or right: [Expected string but got JsonNull],[Expected JsonArray but got JsonNull]],[Expected JsonObject but got JsonNull]
[01:05:29 WARN]:        at com.viaversion.viaversion.libs.mcstructs.converter.model.Result$Error.getOrThrow(Result.java:155)
[01:05:29 WARN]:        at com.viaversion.viaversion.libs.mcstructs.text.serializer.TextComponentCodec.deserialize(TextComponentCodec.java:243)
[01:05:29 WARN]:        at com.viaversion.viaversion.util.SerializerVersion.toComponent(SerializerVersion.java:94)
[01:05:29 WARN]:        at com.viaversion.viaversion.protocols.v1_21to1_21_2.rewriter.BlockItemPacketRewriter1_21_2.lambda$downgradeItemData$20(BlockItemPacketRewriter1_21_2.java:595)
[01:05:29 WARN]:        at com.viaversion.viaversion.api.minecraft.data.StructuredDataContainer.replace(StructuredDataContainer.java:175)
[01:05:29 WARN]:        at com.viaversion.viaversion.protocols.v1_21to1_21_2.rewriter.BlockItemPacketRewriter1_21_2.downgradeItemData(BlockItemPacketRewriter1_21_2.java:582)
[01:05:29 WARN]:        at com.viaversion.viaversion.protocols.v1_21to1_21_2.rewriter.BlockItemPacketRewriter1_21_2.handleItemToServer(BlockItemPacketRewriter1_21_2.java:485)
</code></pre>
</details>

I can update the PR to remove the redundant `getValue() == null` check and refine the `try-catch` if you agree that catching this `JsonParseException` is the right approach to prevent the server from kicking players for clicking GUI items!

kennytv:
Please don't give me AI generated responses, either investigate the issue yourself or I'd ask you to close the PR and create an issue with a reproducible setup instead, otherwise this is a genuine waste of everyone's time. I was asking about *not* "null" strings

EsMokyGame:
I do not speak English and I am using a translator to communicate, so please be a bit more respectful. I am just trying to help fix a bug in the project.

To answer your technical questions directly:
1. You are right, `getValue()` is never null there. 
2. We shouldn't use a `try-catch` because it would hide actually malformed packets that *should* cause a disconnect. 

The issue happening here is the literal string `"null"` being sent by the vanilla client when it interacts with a Shulker Box that somehow got an empty Lock NBT tag (`Lock: ""`) or `"null"`. This happens when players simply click these Shulker Boxes inside their own survival inventory, causing them to be instantly kicked.

kennytv:
Malformed input that would usually disconnect a client (on the original server version) should not be silently ignored in ViaVersion. If adding the empty string and `"null"` check fixes your particular issue, then you shouldn't include any other changes in this PR/remove the try catch

EsMokyGame:
You're totally right. I removed the try-catch so we don't hide other errors.

I just added the check for the empty string and "null" exactly like you suggested, and pushed the changes. Thanks for the help!

kennytv:
~~I can't reproduce this issue with either `null` or `` as the original value when the item is converted to 1.21.11 in the upgrade handler. Being put into a json string, `null` will always be escaped.~~ and empty strings do not matter

common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/BlockItemPacketRewriter1_21_2.java
```diff
@@ -588,7 +588,7 @@ public static void downgradeItemData(final Item item) {
 
             // Back from json in the string tag to plain text
             final StringTag customName = TagUtil.getNamespacedStringTag(itemComponentsTag, "custom_name");
-            if (customName == null) {
+            if (customName == null || customName.getValue().isEmpty() || customName.getValue().equals("null")) {
```
kennytv:
```suggestion
            if (customName == null || customName.getValue().equals("null")) {
```

kennytv (approved):

PR #5005, opened 2026-07-15
fix(protocol): work around 1.21.2 client-side display pitch

### LINKED ISSUE #4907: When using a 1.21.2 or newer client to connect to a 1.20.6 server, there is an issue with incorrect Y-axis rotation on displays.
### `/viaversion dump` Output

 https://dump.viaversion.com/1ec5240150b3e9275acae489df6ea3a295d311e1f4e33ab44647dd2a217f1713

### Server/Client Error

none

### Bug Description

In 1.21.2+ clients connected to a 1.20.6 server, adjusting the Y-axis rotation of item displays does not render correctly when the angle exceeds 90° or -90°. The rotation between 90°180° and -90°-180° fails to display properly.

### Steps to Reproduce

Attempt to rotate an item display by setting its angle using the `setPitch` and `setYaw` methods from a regular entity (via this custom plugin: [ModernShowcase](https://github.com/BrilliantTeam/ModernShowcase).

Then observe the display angle using both a 1.20.6 client and a 1.21.2+ client.

### Expected Behavior

On a 1.20.6 client, the item display renders as expected.
However, on a 1.21.2 or newer client, the display fails to show Y-axis rotations beyond ±90°, making 90°–180° and -90°–-180° rotations visually incorrect or absent.

1.20.6
https://github.com/user-attachments/assets/dff5d906-0760-4674-8545-42c6ceb22422

1.21.2
https://github.com/user-attachments/assets/4d1fbafc-ae98-4ebf-9893-7233b43b77cc

### Additional Server Info

_No response_

### Checklist

RICE0707:
Minecraft 1.21.2 clamps entity pitch to [-90, 90] degrees. For FIXED billboard 
displays that inherit orientation from entity pitch, this breaks rotations 
beyond the clamp range.

We resolve this by storing the raw pitch, calculating the excess angle, and 
pre-multiplying it into the display's left_rotation quaternion (which is unclamped).

- Track display spawn and update pitch in 1.21.2 tracker
- Add helper math to bake clamped-off pitch into left_rotation

Tested and verified working properly on a Paper 1.21.1 server with 
1.21.2 to 26.2 clients.
https://youtu.be/a6zYBgJnkiQ

Fixes #4907


Mukller:
Verified the quaternion math independently — the Hamilton product and the pre-multiplied `rotX(excess) * base` composition are correct, and folding only the clamp overflow while keeping FIXED-billboard gating (with re-bake on billboard transitions) is a clean design. The `appliedLeftRotation` dedup is a nice bandwidth touch.

Two things before this can merge:

1. **Formatting**: several added lines carry trailing whitespace and the `storeEntityPositionRotation` method lost its indent level in this hunk. The repo's spotless/format checks will flag these — worth running the formatter before the next push.

2. **`ADD_ENTITY` byte index**: you read pitch as `wrapper.get(Types.BYTE, 0)`. Since 1.21's `ADD_ENTITY` carries `pitch` *before* `yaw`, this is correct only if no earlier appended handler on this packet has already consumed bytes from the wrapper when yours runs. It holds today given the current registration order, but it silently breaks if someone reorders or adds another `appendClientbound(ADD_ENTITY, ...)` that passthroughs fields first. A short comment stating the assumed read-offset contract (or asserting the expected remaining-readable shape) would protect it.

Also consider covering the spawn-time pitch path in a test if it isn't already — the metadata-update path is exercised by rotations, but a display spawned already pitched beyond ±90° goes through the different code path you added here.

PR #5007, opened 2026-07-16
fix chest_minecart rotation on Minecraft 26+

miko-mak:
After the 1.21.11 version, there was an issue with chest minecarts' rotation. Now the rotation is applied correctly according to the entity's actual position. This visual bug affected **only chest minecarts**. 

The yaw was always off by 180° - represented as +128 (byte angle) in ADD_ENTITY and +180f (float degrees) in ENTITY_POSITION_SYNC.

The fix was made by adding 2 packets into EntityPacketRewriter26_1.java:
**ADD_ENTITY** - corrects the initial rotation when the minecart is spawned.
**ENTITY_POSITION_SYNC** - corrects the rotation on every position update

**Comparison** 
without my fix:

with fix:

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
@@ -545,6 +545,20 @@ public interface ViaVersionConfig extends Config {
     boolean use1_8HitboxMargin();
 
     /**
+     * If enabled, players on 26+ clients will see correct rotation of chest minecarts
+     *
+     * @return true if enabled
+     */
+    boolean isCorrectChestMinecartYaw();
+
+    /**
+     * Returns the rotation offset applied to chest minecarts for 26+ clients on older-version servers.
+     *
+     * @return yaw offset in degrees
+     */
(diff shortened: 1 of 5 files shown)


kennytv (changes requested):
Minecarts now respect facing direction of block states, this isn't the right fix/will break other blocks

miko-mak:
> Minecarts now respect facing direction of block states, this isn't the right fix/will break other blocks

Thanks for the review. Could you clarify what you mean and how this could be fixed instead? My fix always applied the correct rotation in every case I tested.

kennytv:
The block state for minecarts is sent via entity data, where presumably the default chest block state just needs to be mapped to a different one with the opposite facing direction. You can also open an issue instead and we'll fix it for you

miko-mak:
> The block state for minecarts is sent via entity data, where presumably the default chest block state just needs to be mapped to a different one with the opposite facing direction. You can also open an issue instead and we'll fix it for you

Which packet should I listen to? I tried entity set data but it didn’t work

kennytv (changes requested):
It also needs to handle set rot and pos_rot packets

PR #5013, opened 2026-07-28
feat: move removed translations handling into mapping files

### LINKED ISSUE #4892: Move removed translations handling into mapping files.
### Problem Description

Handling removed translations in VV should be moved into mapping files, similar to how everything else is mapped in newer versions. 

### Solution Description

Add support for adding/removing translation keys in mapping files, this would also make the translation-mappings json in VB obsolete.

### Alternatives

Continue as is 

### Additional Info

_No response_

bejiihiu:
Closes #4892.

Moves hardcoded translation key mappings out of ComponentRewriter Java classes into `mapping-lang-*.txt` resource files, consistent with how `mapping-lang-1.12-1.13.txt` already works for the 1.12→1.13 transition.

**Problem:** Every version transition that removes or renames translation keys had the mappings hardcoded as `HashMap.put()` calls in their ComponentRewriter — inconsistent with every other mapping type (blocks, items, sounds, entities, etc.) which are all loaded from data files.

**Changes:**

1. Added `getTranslationMappings()` to the `MappingData` interface and `MappingDataBase`. The base `loadExtras()` now auto-discovers `mapping-lang-{unmappedVersion}to{mappedVersion}.txt` by naming convention, skipping silently if absent.

2. Created 7 mapping files under `assets/viaversion/data/`:
   - `mapping-lang-1.15to1.16.txt` — 34 entries (attribute renames, block/item display names)
   - `mapping-lang-1.16.2to1.17.txt` — 10 entries (worldborder, replaceitem, debug, locate)
   - `mapping-lang-1.17to1.18.txt` — 41 entries (scoreboard, 39 removed biome names)
   - `mapping-lang-1.18to1.19.txt` — 9 entries (locate, placefeature renames)
   - `mapping-lang-1.19to1.19.3.txt` — 4 entries (locate poi/biome, disconnect messages)
   - `mapping-lang-1.21to1.21.2.txt` — 8 entries (drop, advancement, function, fillbiome, publish)
   - `mapping-lang-1.21.11to26.1.txt` — 2 entries (time set/query)

3. Simplified 7 ComponentRewriters to populate from `protocol.getMappingData().getTranslationMappings()` instead of hardcoded maps or switch statements:
   - `ComponentRewriter1_16` (−34 hardcoded entries)
   - `ComponentRewriter1_17` (−10 entries)
   - `ComponentRewriter1_18` (−41 entries)
   - `ComponentRewriter1_19` (−9 entries)
   - `ComponentRewriter1_19_3` (switch → map lookup)
   - `ComponentRewriter1_21_2` (switch → map lookup)
   - `ComponentRewriter26_1` (switch → map lookup)

4. Updated `MappingData1_19` and `MappingData26_1` to call `super.loadExtras(data)` so translation files load alongside their existing extras.

**Format** follows the existing `mapping-lang-1.12-1.13.txt` convention:
```
old_key=new_value
```
`new_value` is either a renamed key or inlined English text for removed keys. Lines starting with `#` are treated as comments.

**Backward compatibility:** Zero behavioral change — identical mappings are loaded from files instead of Java code. `\n` escape sequences in values are converted to actual newline characters on load.

**Validation:**
- Full `./gradlew build` — **BUILD SUCCESSFUL**
- All tests pass (`:viaversion-common:test`, `:viaversion-api:test`)


kennytv:
This approach is worse than just leaving it as is

PR #5016, opened 2026-07-29
fix: skip empty items in convertItemList to prevent air serialization…

1wairesd:
… errors

Previously, convertItemList wrote all slots including null/empty items to NBT as "minecraft:air" with a Slot tag. On 1.21+ servers, this caused TypedEntityData serialization errors because air is no longer valid in item compound data. Empty items are already filtered out when reading back via itemFromTag (returns empty → filtered by isEmpty check), so skipping them during write is safe and changes no behavior.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/StructuredDataConverter.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_20_3to1_20_5/rewriter/StructuredDataConverter.java
@@ -798,6 +798,9 @@ public final class StructuredDataConverter {
         final ListTag<CompoundTag> itemsTag = new ListTag<>(CompoundTag.class);
         for (int i = 0; i < items.length; i++) {
             final Item item = items[i];
+            if (item == null || item.isEmpty()) {
+                continue;
+            }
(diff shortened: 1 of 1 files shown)


Mukller:
The fix itself looks right: `putByte("Slot", (byte) i)` uses the original loop index (line 804), so per-item slot tags stay correct after skipping, and vanilla's `ContainerHelper`-style readers key off `Slot`, not position.

One behavioral change worth being deliberate about: previously the emitted list always had `size() == items.length` (one entry per slot, air included). After this change the list can be shorter than the container with gaps implied by `Slot`. Anything in the ecosystem that reads these tags *positionally* instead of via `Slot` — some plugin-side NBT wrappers do — will now see shifted inventories where they previously saw exact 1:1 layouts. That may be perfectly acceptable (arguably more correct NBT), but it's observable to consumers outside this repo.

Suggestion: make sure test coverage includes a **sparse** array — e.g. empty slot 0, item in slot 1 — asserting both the surviving entry and its `Slot` value. A dense-array test can't distinguish "skipped" from "shifted", which is exactly the failure mode above.

kennytv:
Do you have example items for this? Pretty sure old clients have/need air serialized explicitly in some containers

PR #5019, opened 2026-08-05
Allow servers predating data components to declare them via legacy nbt

### LINKED ISSUE #5018: Allow servers predating data components to declare them via legacy nbt
### Problem Description

`BlockItemPacketRewriter1_20_5#toStructuredItem` converts legacy item nbt into the data components that have a legacy equivalent: `display.*`, `Damage`, `CustomModelData`, `Trim`, `SkullOwner`, `AttributeModifiers` and so on. That conversion is complete for everything that ever existed as nbt.

Components introduced after 1.20.5 have no legacy ancestor, so servers older than that cannot express them at all. There is no key to convert from, and unrecognised nbt ends up in `minecraft:custom_data`, which the client only surfaces in tooltips.

`minecraft:equippable` is the clearest case. Its `asset_id` is a plain identifier resolved client-side against `assets/<namespace>/equipment/<path>.json`, with no registry entry and no config-phase sync, unlike `trim_pattern`/`trim_material`. It is the only way to give a modern client per-item armor textures, and a 1.8 backend has no route to it. The closest workaround is abusing `Trim`, which crosses unmodified but caps out at the number of pattern/material combinations.

To be upfront about the framing: this is not restoring pre-existing behaviour of an old client, it is letting an old server opt into a modern component. I understand that is a different category from most accepted changes, which is why I am asking before opening a PR.

### Solution Description

An opt-in `VV|components` sub-tag of the item's legacy nbt, holding data components keyed by their vanilla identifier, using the vanilla field names:

```
tag: {
    "VV|components": {
        equippable: {
            slot: "chest",
            asset_id: "example:my_armor"
        }
    }
}
```

A `LegacyComponentConverter`, the inverse of the existing `StructuredDataConverter`, placed in the protocol that first supports the components it converts. For `equippable` that is `v1_21to1_21_2`, which is also the gate: that protocol only runs for servers older than 1.21.2, so no explicit `serverProtocolVersion` check is needed.

Points that I think matter for whether this is acceptable:

- Opt-in via `convert-legacy-component-nbt`, default `false`. When disabled the cost is one boolean field read per item.
- The serverbound path needs no new code. `EQUIPPABLE1_21_2` is already in `BlockItemPacketRewriter1_21_2.NEW_DATA_TO_REMOVE`, so `downgradeItemData` strips it before the item reaches `toOldItem`. The backend server never sees the component.
- The nbt is deliberately left in `custom_data`, since that tag is the original legacy nbt that `toOldItem` hands back to the server verbatim. Removing it would make the server see the item lose data as soon as the client moved it.
- The existing `replaceKey` chain carries `EQUIPPABLE1_21_2` to `EQUIPPABLE1_21_5` to `EQUIPPABLE1_21_6`, so there is no per-version work.
- The `VV|` prefix follows the convention in `ItemRewriter#nbtTagName()`, and keeps the tag from colliding with anything a plugin may already store on items.
- Converters are a table, so supporting further components is one line plus a parse method. I kept it to `equippable` rather than guessing which others are wanted.

I have this implemented, unit tested (9 tests) and verified in-game on a live 1.8 backend. Happy to
open the PR, adjust the shape, or drop it, whichever you prefer.

### Alternatives

- A `Provider` hook, so implementations supply the components themselves. Smaller in-tree, and there is precedent in `SignableCommandArgumentsProvider` (#3493), but it pushes the legacy nbt contract onto every consumer and adds API surface for a single caller.
- Adding `equippable` to the `toStructuredItem` allowlist. Wrong layer: the key does not exist in the 1.20.5 component set and the container's id lookup is in 1.20.5 space there.
- Reusing `VV|DataComponents`. That is ViaBackwards' backup format for data lost on the way down, with ad-hoc per-component shapes, and it is write-only in this repository, since the restore lives in ViaBackwards. Conflating the two looked likely to collide when both are in the pipeline.
- Maintaining a fork. Works, and I will do that if this is out of scope, but every server owner wanting this would need their own.

### Additional Info

Verified on a 1.8.9 backend with a custom armor asset: the component reaches the client, renders on the player model, and the item survives inventory moves and drops with its nbt intact on the server side.

Applies to any backend older than 1.21.2. `minecraft:equippable` was added in 1.21.2, so `v1_21to1_21_2` is the earliest protocol that can carry it. No registry synthesis is required for `asset_id`: it is `Types.OPTIONAL_STRING` on the wire in all three `Equippable` variants and is resolved entirely by the client's resource pack.

IceGames23:
Closes #5018

## Problem

`BlockItemPacketRewriter1_20_5#toStructuredItem` converts legacy item nbt into the data components that
have a legacy equivalent: `display.*`, `Damage`, `CustomModelData`, `Trim`, `SkullOwner`,
`AttributeModifiers` and so on. That conversion is complete for everything that ever existed as nbt.

Components introduced after 1.20.5 have no legacy ancestor, so servers older than that cannot express
them at all. There is no key to convert from, and unrecognised nbt ends up in `minecraft:custom_data`,
which the client only surfaces in tooltips.

## Solution

An opt-in `VV|components` sub-tag of the item's legacy nbt, holding data components keyed by their
vanilla identifier and using the vanilla field names:

```
tag: {
    "VV|components": {
        equippable: {
            slot: "chest",
            asset_id: "example:my_armor"
        }
    }
}
```

`LegacyComponentConverter` is the inverse of the existing `StructuredDataConverter` and lives in the
protocol that first supports the components it converts. That also gates it: `v1_21to1_21_2` only runs
for servers older than 1.21.2, so no explicit `serverProtocolVersion` check is needed.

Only `minecraft:equippable` is converted so far. The converters are a table, so supporting further
components is one line plus a parse method. I kept it to one rather than guessing which others are
wanted.

## Notes for review

- **Opt-in.** `convert-legacy-component-nbt` defaults to `false`. When disabled the cost is one boolean
  field read per item.
- **The serverbound path needs no new code.** `EQUIPPABLE1_21_2` is already in `NEW_DATA_TO_REMOVE`, so
  `downgradeItemData` strips it before the item reaches `toOldItem`. The backend server never sees the
  component.
- **The nbt is deliberately left in `custom_data`.** That tag is the original legacy nbt handed back to
  the server verbatim by `toOldItem`'s short-circuit. Removing it would make the server see the item
  lose data as soon as the client moved it. `testComponentNbtIsLeftInCustomData` covers this.
- **No per-version work.** The existing `replaceKey` chain carries `EQUIPPABLE1_21_2` to
  `EQUIPPABLE1_21_5` to `EQUIPPABLE1_21_6`.
- **Sounds use direct holders**, so later protocols' sound rewriting passes them through unchanged
  rather than remapping a numeric id out of this protocol's id space.
- `allowed_entities` is not supported, since it would need entity registry resolution at this point.

## Tests

9 tests in `LegacyComponentConverterTest`, covering the defaults, explicit fields, the nbt surviving in
`custom_data`, and the rejection paths:

- unknown slot, which matters because `EnumType#idFromName` falls back to the first entry, so a typo
  would otherwise silently equip to `mainhand`
- invalid `asset_id`, which would otherwise throw in the 1.21.5+ hash codec, far from where it was set
- missing slot, unknown component, unnamespaced tag, and items without `custom_data`

Also verified in-game on a 1.8.9 backend with a 1.21.11 client: the component reaches the client and
renders on the player model, using an `asset_id` from a resource pack that renders identically when
given directly on a vanilla 1.21.11 client.

## Alternatives considered

- **A `Provider` hook** so implementations supply the components themselves. Smaller in-tree, with
  precedent in `SignableCommandArgumentsProvider` (#3493), but it pushes the legacy nbt contract onto
  every consumer and adds API surface for a single caller.
- **Adding `equippable` to `toStructuredItem`'s allowlist.** Wrong layer: the key does not exist in the
  1.20.5 component set and the container's id lookup is in 1.20.5 space there.
- **Reusing `VV|DataComponents`.** That is ViaBackwards' backup format for data lost on the way down,
  with ad-hoc per-component shapes, and it is write-only in this repository since the restore lives in
  ViaBackwards. Conflating the two looked likely to collide when both are in the pipeline.

### DIFF
--- a/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
@@ -550,7 +550,7 @@ public interface ViaVersionConfig extends Config {
      *
      * @return true if enabled
      */
-    boolean convertLegacyComponentNbt();
+default boolean convertLegacyComponentNbt() { return false; }
 
     /**
      * If enabled, ViaVersion will send the native client version to the server on connect via a plugin message.
--- a/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
+++ b/api/src/main/java/com/viaversion/viaversion/api/configuration/ViaVersionConfig.java
@@ -545,6 +545,14 @@ public interface ViaVersionConfig extends Config {
     boolean use1_8HitboxMargin();
 
     /**
+     * If enabled, data components declared by servers older than the version that added them are read from a
+     * {@code VV|components} sub-tag of the item's legacy nbt and sent to the client.
+     *
+     * @return true if enabled
+     */
+    boolean convertLegacyComponentNbt();
+
+    /**
      * If enabled, ViaVersion will send the native client version to the server on connect via a plugin message.
      *
      * @return true if enabled
--- a/common/src/main/java/com/viaversion/viaversion/configuration/AbstractViaConfig.java
+++ b/common/src/main/java/com/viaversion/viaversion/configuration/AbstractViaConfig.java
@@ -101,6 +101,7 @@ public class AbstractViaConfig extends Config implements ViaVersionConfig {
     private boolean cancelSwingInInventory;
     private int maxErrorLength;
     private boolean use1_8HitboxMargin;
+    private boolean convertLegacyComponentNbt;
     private boolean sendPlayerDetails;
     private boolean sendServerDetails;
 
@@ -177,6 +178,7 @@ public class AbstractViaConfig extends Config implements ViaVersionConfig {
         fix1_21PlacementRotation = getBoolean("fix-1_21-placement-rotation", true);
         cancelSwingInInventory = getBoolean("cancel-swing-in-inventory", true);
         use1_8HitboxMargin = getBoolean("use-1_8-hitbox-margin", true);
+        convertLegacyComponentNbt = getBoolean("convert-legacy-component-nbt", false);
         sendPlayerDetails = getBoolean("send-player-details", true);
         sendServerDetails = getBoolean("send-server-details", true);
         packetTrackerConfig = loadRateLimitConfig(getSection("packet-limiter"), "%pps", 1);
@@ -650,6 +652,11 @@ public class AbstractViaConfig extends Config implements ViaVersionConfig {
     @Override
     public boolean use1_8HitboxMargin() {
         return use1_8HitboxMargin;
+    }
+
+    @Override
+    public boolean convertLegacyComponentNbt() {
+        return convertLegacyComponentNbt;
     }
 
     @Override
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/BlockItemPacketRewriter1_21_2.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_21to1_21_2/rewriter/BlockItemPacketRewriter1_21_2.java
@@ -23,6 +23,7 @@ import com.viaversion.nbt.tag.IntArrayTag;
 import com.viaversion.nbt.tag.ListTag;
 import com.viaversion.nbt.tag.StringTag;
 import com.viaversion.nbt.tag.Tag;
+import com.viaversion.viaversion.api.Via;
 import com.viaversion.viaversion.api.connection.UserConnection;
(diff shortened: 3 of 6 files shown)


kennytv:
Not interested in reviewing slop or an unnecessarily wordy PR description, sorry. This comes with a few issues

IceGames23:
My english is bad, I need help to write this PR description, but what's the exact issue on the code? @kennytv

PR #5022, opened 2026-08-06
Synchronize EntityTrackerBase entity map to prevent concurrent-access crash

XEXModz:
Ran into a crash where the `entities` map in `EntityTrackerBase` gets touched from more than one thread and corrupts during a rehash. It throws an out of bounds exception that turns into an `EncoderException` and kicks the player.

```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 65
    at it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap...
    at ...EntityTrackerBase...
    (becomes) io.netty.handler.codec.EncoderException
```

It showed up for me running Via alongside GrimAC, since it bundles PacketEvents which can hit the tracker off the main packet thread. In my case only the older (pre-26.2) clients going through translation got kicked.

The fix just wraps the map with `Int2ObjectMaps.synchronize(...)` so access is serialized. I know the tracker is normally treated as single threaded, so this might be considered the other plugin's problem, but a synchronized wrapper is cheap and it kills a crash that was a pain to track down.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/data/entity/EntityTrackerBase.java
+++ b/common/src/main/java/com/viaversion/viaversion/data/entity/EntityTrackerBase.java
@@ -30,6 +30,7 @@ import com.viaversion.viaversion.api.type.Type;
 import com.viaversion.viaversion.util.Key;
 import com.viaversion.viaversion.util.KeyMappings;
 import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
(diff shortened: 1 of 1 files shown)


kennytv:
Except for one call on 1.8 servers in our own code, this map should never be accessed from anywhere but a netty thread (not the main thread either) and it'd break much more than just this specific access

PR #5025, opened 2026-08-12
Use instanceof check in ConfigSection#getSerializedComponent

Smorki:
`getSerializedComponent` casts the value to `String` directly without checking the type. Other getter methods in the same class (`getString`, `getInt`, etc.) already use `instanceof` to guard against wrong types.

If a config has a non-String value for the key, this method would throw `ClassCastException`. Aligned it with the pattern used by the other getters.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/util/ConfigSection.java
+++ b/common/src/main/java/com/viaversion/viaversion/util/ConfigSection.java
@@ -117,12 +117,10 @@ public class ConfigSection {
     }
 
     public @Nullable JsonElement getSerializedComponent(final String key) {
-        final Object o = this.values.get(key);
-        if (o != null && !((String) o).isEmpty()) {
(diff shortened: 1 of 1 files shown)


Mukller:
Verified against the current codebase: `getString`, `getInt`, `getDouble`, `getListSafe`, and `getSection` all already use this exact `instanceof final` pattern, so both the consistency claim and the language level check out.

One behavioral note: the only caller is `AbstractViaConfig` (`resource-pack-1_17-prompt`), where a wrong-typed value previously failed loudly with `ClassCastException` at load time and will now silently yield a null prompt. That's consistent with this class's getter convention, so this looks good to me — though since `getListSafe` threads an `invalidValueMessage` through for exactly this situation, a debug-level log when a non-String value gets dropped might make misconfigured prompt messages easier to diagnose.

kennytv (approved):

Smorki:
thanks

PR #5027, opened 2026-08-16
Skip identity encode copies

bridgelol:
`shouldTransformPacket()` is just `UserConnection.active`. ViaBackwards clients stay active for the whole session, so the encoder `ByteBufUtil.copy`s every outbound packet even when nothing in the pipeline will touch it.

This peeks the packet id first. If no protocol has a mapping for that id and none override `transform`, the buffer is retained as-is. Protocols that do override `transform` (ViaBackwards config-bridge stuff, `Protocol1_20To1_20_2`, …) keep the old path via `maySkipUnregisteredPackets()` defaulting to false.

Native-protocol clients already flip `active` off after login. They also replace the encode/decode codecs with a named passthrough so `sendRawPacket` / `JoinListener` still resolve `via-encoder`.

`EntityType.isOrHasParent` is cached for enum types (metadata filter hot path).

API additions are default methods only, so ViaBackwards/ViaRewind should keep compiling.


kennytv:
Please provide benchmarks and real world performance numbers / reasons for these changes, especially the entity parent check (after replacing the arrays with bitsets, because that storage format seems unnecessarily wasteful and will keep growing exponentially) and full additional pipeline traversal of long pipelines just for skipping a native pooled copy

bridgelol:
The extra pipeline walk was the wrong trade, yeah.

On a long pipe (20 protocols) an identity-id check was ~44ns, same ballpark as the pooled copy it was trying to skip (~36-41ns for 64-1500B). So for ViaBackwards-length pipes that version of the skip was a wash, and it added work on packets that still have to transform.

Pushed a follow-up that builds a registered-id BitSet when the pipeline changes:

- If any protocol overrides `transform()` (`maySkipUnregisteredPackets() == false`, e.g. 1.20->1.20.2, ViaBackwards config-bridge stuff) the pipe is marked non-passthrough. Encode does the same `active` check as before. No id peek, no walk.
- Mapping-only pipes (ViaVersion-only / adjacent versions) peek the id and test the bitset (~2ns) instead of copying.

That's the case this is actually for. Non-native clients stay `active=true` for the whole session, so every outbound packet currently hits `ByteBufUtil.copy` on the event loop. Adjacent-version pipes have a lot of packets with no mapping. Native clients already flip `active` off after login; the named passthrough is only so `sendRawPacket` / `JoinListener` still resolve `via-encoder`.

Entity parent cache: dropped. Walk is ~4ns, a packed bitset was ~3ns. `boolean[N][N]` was ~880KiB across the entity enums and still O(N²); even bitsets are ~142KiB for a 1ns delta. Not worth it.

PR #5031, opened 2026-08-23
Fix ByteBuf leak on serverbound config queue overflow in Configurationstate

Smorki:
In `ConfigurationState#addServerboundPacketToQueue`, `toQueuedPacket` creates a fresh `ByteBuf` via `Unpooled.buffer()`.

If the serverbound packet count or byte limit is exceeded, the method calls `user.disconnect()` and returns early. Since the `QueuedPacket` is never added to `packetQueue`, its underlying buffer is not released during `onRemove()` or queue drain, leaking the allocated `ByteBuf`.

This PR releases the buffer before returning on disconnect.


kennytv:
It's an unpooled heap buffer, this doesn't really matter

Smorki:
Fair point on GC handling unpooled heap buffers — feel free to close this if its not worth merging!

Im also opening another PR to memoize `FullMappingsBase#inverse` to reduce allocations during item rewriting which should be more impactful

PR #5032, opened 2026-08-23
Memoize FullMappingsBase.inverse() using thread-safe double-checked locking

Smorki:
This PR memoizes the inverse mapping instance inside `FullMappingsBase` using thread-safe double-checked locking so it is lazily initialized once and reused.


Smorki:
@kennytv

kennytv:
The inverse itself is already cached so all this does is synchronize to avoid a single wrapper object. No more slop please

PR #5037, opened 2026-08-30
Fix console errors from sending stats removed in 1.11.1 to 1.11.1+ clients

LeonTG:
The errors didn't break anything, but they were annoying and spammed often enough for me to fix it upstream

**Errors:**
image

**Source of stats removal:**
https://minecraft.wiki/w/Java_Edition_1.11.1#:~:text=Removed%20the%20%27Junk%20Fished%27%20and%20%27Treasure%20Fished%27%20statistics


common/src/main/java/com/viaversion/viaversion/protocols/v1_11to1_11_1/Protocol1_11To1_11_1.java
```diff
@@ -33,6 +36,29 @@ public Protocol1_11To1_11_1() {
     @Override
     protected void registerPackets() {
         itemRewriter.register();
+
+        registerClientbound(ClientboundPackets1_9_3.AWARD_STATS, wrapper -> {
+            Object2IntMap<String> filteredStats = new Object2IntOpenHashMap<>();
+            int size = wrapper.read(Types.VAR_INT);
```
kennytv: use wrapper.set to change the size instead of putting everything into a new map

kennytv (changes requested):

common/src/main/java/com/viaversion/viaversion/protocols/v1_11to1_11_1/Protocol1_11To1_11_1.java
LeonTG: Better now?

kennytv (approved):

PR #5039, opened 2026-09-01
Fix synthetic timeline tag bindings

EvModder:
Adds the missing timeline tag bindings when translating 1.21.9 configuration data to 1.21.11.

Bindings are appended to existing tag packets, or synthesized before `finish_configuration` when a server sends none.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_21_9to1_21_11/Protocol1_21_9To1_21_11.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_21_9to1_21_11/Protocol1_21_9To1_21_11.java
@@ -66,6 +66,7 @@ import com.viaversion.viaversion.rewriter.TagRewriter;
 import com.viaversion.viaversion.rewriter.block.BlockRewriter1_21_5;
 import com.viaversion.viaversion.rewriter.text.NBTComponentRewriter;
 import com.viaversion.viaversion.util.Key;
+import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


kennytv:
What exactly is this supposed to fix?

EvModder:
Mainly, showing wrong clock time in F3 (debug screen) or mods like MiniHud.

I noticed this while playing on a Multiplayer server running an older version, mainly because I live in the end dimension, and rely on the "synthetic" clock (the F3 debug or MiniHud display line) to know when I can trade with my villagers lol.

Edit:
Ah wait, I see I replied on the wrong PR, I was thinking this was #5040
*This* PR is not strictly a prerequisite (for players on 1.21.11+), but is a distinct/adjacent protocol transition (timeline registry tags); 1.21.9 -> 1.21.11, whereas #5040 handles clock protocol changes 1.21.11 -> 26.1+
Technically in my case, I'm playing on 26.2 client <-> 1.21.4 server

kennytv:
That explains the other PR but still not this one

EvModder:
It's two-step with the prior PR.

This fixes “unbound `minecraft:timeline` tags” disconnects, by supplying missing timeline tag bindings at the 1.21.9 → 1.21.11 translation step; #5040 fixes clock values at the 1.21.11 → 26.1 step. My 26.2 client → 1.21.4 server connection passes through both.

P.S.:
Sorry for the slow reply; I've been traveling and mostly without internet.

PR #5041, opened 2026-09-03
Rewrite teleport reply position to fix riding teleport loop on 1.8 servers

### LINKED ISSUE #4748: [1.8 -> 1.16.5] Packet spam when riding entities leads to clients being kicked
### `/viaversion dump` Output

https://dump.viaversion.com/ba1aa1a6eac3a215ca632949a4dd3484c499489d0c5e5c07e4bb1897e38deee6

### Server/Client Error

Kicked for "You are sending too many packets" eventually.

### Bug Description

When connecting modern Minecraft clients to a Paper 1.8.8 server via ViaVersion, clients sometimes get kicked by for “You are sending too many packets!”. Packet captures reveal the client spamming Move Player PosRot and Accept Teleportation packets at extremely high rates (multiple packets per millisecond) while the server is sending massive floods of "Player Position" packets, seemingly creating an endless pingpong.

S -> C
Image

C -> S
Image

The packets have repeated positions/rotation vectors, and Accept Teleportation packets have ID 0.

This occurs seemingly random when mounting entities (horses, etc.).

The issue appears to be specific to legacy 1.8 servers; testing on Paper 1.16.5 does not reproduce the behavior.

### Steps to Reproduce

Steps to Reproduce:

- Set up Paper 1.8.8 server.
- Install ViaVersion 5.6.0 (no ViaBackwards needed to reproduce).
- Connect using a modern client (1.21.11 in this case)
- Mount and dismount an entity (e.g., horse) repeatedly until it appears behaving weirdly (check video)

Observe packet spam in a network capture or get kicked for too many packets.

https://www.youtube.com/watch?v=sICLeZ7CxN8

### Additional Server Info

- Paper 1.8.8 (git-CurrySpigot-"280cd42")
- Tested with **and** without proxy (bungee)
- ViaVersion 5.6.0
- No other packet-modifying plugins required to reproduce
- Tested with and without additional plugins; issue persists on clean setup

### Checklist

Packsolite:
Fixes #4748

Since 1.21.2, the client keeps its own position when it receives a teleport while riding and replies with that position instead of the teleport position. As from the  [protocol docs](https://minecraft.wiki/w/Java_Edition_protocol/Packets#Synchronize_Player_Position):
> If the player is riding a vehicle, this packet has no effect, but both the [Confirm Teleportation](https://minecraft.wiki/w/Java_Edition_protocol/Packets#Confirm_Teleportation) and [Set Player Position and Rotation](https://minecraft.wiki/w/Java_Edition_protocol/Packets#Set_Player_Position_and_Rotation) packets are still sent.

An 1.8 server, however, expects the reply at the teleport position and treats anything else as an "illegal" move and endlessly corrects the player. Each correction is another teleport, so client and server ping-pong until the client is kicked ("You are sending too many packets"). Clients before 1.21.1 applied the teleport even while riding, so this never happened before.

This PR implements a fix by remembering the position of every teleport sent to the client and rewriting the client's reply to it. This matches pre-1.21.2 behavior of the client.

Reproduced on Paper 1.8.8 with a 26.2 client (packet capture + server-side teleport-event log showing ~250 corrections/s until the kick, see #4748). With this change the loop no longer occurs and mount/dismount behaves normally.

Disclosure: Part of the research and implementation of this PR were AI assisted. I've reviewed and verified the results and any changes made.

### DIFF
--- a/common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/storage/TeleportPositionStorage.java
+++ b/common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/storage/TeleportPositionStorage.java
@@ -1,0 +1,50 @@
+/*
+ * This file is part of ViaVersion - https://github.com/ViaVersion/ViaVersion
+ * Copyright (C) 2016-2026 ViaVersion and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.viaversion.viaversion.protocols.v1_8to1_9.storage;
+
+import com.viaversion.viaversion.api.connection.StorableObject;
+import com.viaversion.viaversion.api.minecraft.Vector3d;
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+/**
+ * Positions of the teleports most recently sent to the client.
+ * Since 1.21.2, clients keep their own position when receiving a teleport while riding and reply
+ * with that instead of the teleport position. A 1.8 server expects the reply at the teleport
+ * position, treats anything else as an illegal move, and endlessly corrects the player.
+ * See https://github.com/ViaVersion/ViaVersion/issues/4748
+ */
+public class TeleportPositionStorage implements StorableObject {
+    private static final int MAX_PENDING_POSITIONS = 64;
+
+    private final Deque<Vector3d> pendingPositions = new ArrayDeque<>();
+
+    public void addPendingPosition(Vector3d position) {
+        if (this.pendingPositions.size() >= MAX_PENDING_POSITIONS) {
+            this.pendingPositions.poll();
+        }
+        this.pendingPositions.add(position);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/storage/TeleportPositionStorage.java
```diff
@@ -0,0 +1,50 @@
...
+        if (this.pendingPositions.size() >= MAX_PENDING_POSITIONS) {
+            this.pendingPositions.poll();
+        }
+        this.pendingPositions.add(position);
+    }
+
+    /**
+     * Returns and removes the oldest pending teleport position, or null if none is pending.
+     */
+    public Vector3d pollPendingPosition() {
```
kennytv: @ nullable annotation

common/src/test/java/com/viaversion/viaversion/common/storage/TeleportPositionStorageTest.java
```diff
@@ -0,0 +1,79 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.viaversion.viaversion.common.storage;
+
+import com.viaversion.viaversion.api.minecraft.Vector3d;
+import com.viaversion.viaversion.protocols.v1_8to1_9.storage.TeleportPositionStorage;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class TeleportPositionStorageTest {
```
kennytv: These look pointless, remove the test class

kennytv (changes requested):
The polling feels like it'll break on clients/servers with high latency, depending on where Via is installed? accept teleportation was added specifically made because of that

Packsolite:
> The polling feels like it'll break on clients/servers with high latency, depending on where Via is installed? accept teleportation was added specifically made because of that

Can you elaborate please? In which scenario would it break?

kennytv:
The incoming player movement can be behind what the server is sending so you rewrite one movement to be "correct" and the next one is suddenly dozens of blocks away again

PR #5043, opened 2026-09-11
Fix original client version in Velocity player details

chicoferreira:
Currently, velocity’s `vv:proxy_details` payload reports the intermediate protocol used by Velocity in `version` and `versionName`, instead of the original player’s client version when a frontend translation occurs.

This makes, for example, a 1.8.9 client connecting through Velocity with modern forwarding enabled (which requires ViaVersion to translate its protocol to 1.13) get reported as 1.13 in `vv:proxy_details`.

This PR fixes the player version sent to the backend by reading it from the frontend connection, while continuing to send the server version to the client using information from the backend connection.


kennytv (approved):

florianreuth:
We shouldn't just send the payload twice, this looses the information which connection is meant and backend plugins can't deal with it properly. We would either need a better fix or have some indication inside the payload.

chicoferreira:
> We shouldn't just send the payload twice, this looses the information which connection is meant and backend plugins can't deal with it properly. We would either need a better fix or have some indication inside the payload.

The diff might be confusing, but `sendConnectionDetails` already calls both `sendPlayerDetails` and `sendServerDetails` internally. This PR simply makes those calls separately, so it doesn’t send any additional payloads.

The backend still receives one player-details payload, now using the original client version from the frontend connection. The server details sent to the Minecraft client remain unchanged.

florianreuth:
Ah yeah, seems fine. Thank you :)

PR #5044, opened 2026-09-14
Validate lengths read from buffers before allocating

alexec0de:
## Problem

Several `Type#read` implementations use a length read straight from the buffer to size an
allocation *before* reading any of the elements it describes. Because nothing is read first,
the usual "the buffer runs out" bound never kicks in, and a few bytes on the wire turn into an
arbitrarily large allocation.

Three separate cases:

1. **Palette length** — `PaletteType1_18`, `ChunkSectionType1_13`, `ChunkSectionType1_16` pass the
   length into `DataPaletteImpl`, which forwards it to `new IntArrayList(initialSize)`.
2. **Block entity count** — `ChunkType1_18`, `ChunkType1_20_2`, `ChunkType1_21_5` pass it to
   `new ArrayList<>(length)`.
3. **Element count vs. byte count** — `LongArrayType` and `IntArrayType` sanity-check
   `buffer.isReadable(length)`, but then allocate `length * 8` / `length * 4` bytes. A packet of
   N readable bytes can legally declare N longs, so the check is off by a factor of 8.

Additionally, `ByteBuf#isReadable(int)` is `writerIndex - readerIndex >= numBytes`, which is `true`
for a negative argument, so a negative length reached the allocation and produced a
`NegativeArraySizeException` instead of a clear error.

### Reproduction

A six-byte buffer (`bitsPerBlock = 4`, `paletteLength = 0x7FFFFFFF`) fed to
`Types.CHUNK_SECTION1_16.read`:

```
java.lang.OutOfMemoryError: Requested array size exceeds VM limit
	at it.unimi.dsi.fastutil.ints.IntArrayList.initArrayFromCapacity(IntArrayList.java:85)
	at it.unimi.dsi.fastutil.ints.IntArrayList.<init>(IntArrayList.java:94)
	at com.viaversion.viaversion.api.minecraft.chunks.DataPaletteImpl.<init>(DataPaletteImpl.java:51)
	at com.viaversion.viaversion.api.minecraft.chunks.ChunkSectionImpl.<init>(ChunkSectionImpl.java:46)
	at com.viaversion.viaversion.api.type.types.chunk.ChunkSectionType1_16.read(ChunkSectionType1_16.java:55)
```

These are all clientbound reads, so on a server-side install the bytes come from the server
itself. The exposure is on platforms built on this module where the remote server is not trusted
(ViaProxy, ViaFabric, ViaForge), where this is a remote OOM from a handful of bytes.

## Changes

- Added `Limit#initialCapacity`, a non-throwing clamp for lengths used as allocation hints, next to
  the existing `Limit#max`.
- Palette lengths are clamped to `1 << bitsPerValue` for the pre-allocation only. Indexes are
  `bitsPerValue` wide, so a well-formed palette never holds more entries than that. The read loop
  still consumes the declared length and fails on the buffer as before — this keeps the existing
  tolerant behaviour of these methods, which already sanitise a malformed `bitsPerValue` rather
  than throwing.
- Block entity counts are clamped the same way; the list still grows as needed.
- `LongArrayType` and `IntArrayType` now compare against `readableBytes / Long.BYTES` and
  `readableBytes / Integer.BYTES`. Dividing the readable bytes rather than multiplying the length
  avoids overflowing the `int`, and `readableBytes()` is read once so the hot path gains no extra
  call.
- Negative lengths are rejected in `ArrayType`, `ByteArrayType`, `IntArrayType`, `LongArrayType` and
  `VarIntArrayType`, with the length and readable bytes in the message.
- `VarIntType#readPrimitive` checks the byte count *before* reading, instead of after. Previously a
  sixth byte was consumed and its `<< 35` silently wrapped to `<< 3` before the exception was
  thrown. Five-byte VarInts still parse.

No behaviour change for well-formed data.

## Testing

`LengthValidationTest` covers each case plus a control test that valid arrays still round-trip.
Against the unpatched code:

| Test | before | after |
| --- | --- | --- |
| `testOversizedPaletteIsNotPreAllocated` | `OutOfMemoryError` | pass |
| `testLongArrayLengthCountsBytes` | fail | pass |
| `testIntArrayLengthCountsBytes` | fail | pass |
| `testNegativeLengthsRejected` | fail | pass |
| `testVarIntDoesNotConsumeSixthByte` | fail | pass |
| `testValidArraysStillRead` | pass | pass |


kotikstasika:
**okak**

kennytv:
There's no real point in validating clientbound packets. A server gains nothing from making their own players or server crash, but also has many more and better ways of achieving that than this

PR #5045, opened 2026-09-15
Fix death smoke not sending for 1.17+ clients

LeonTG:
In 1.16.5 and below, the smoke displayed at the end of an entitys death animation was handled client side

In 1.17 and later, it is now sent as an `ENTITY_EVENT` packet when the server determines the death animation is complete

This PR makes 1.17+ clients on 1.16.5 and below servers also get said `ENTITY_EVENT` packet when the death animation is complete

I only implemented this for Bukkit platform, cuz I have no idea how to do it for the other platforms, that can be someone elses problem


kennytv (changes requested):
This would presumably be better by responding to entity events on a protocol level instead

LeonTG:
> This would presumably be better by responding to entity events on a protocol level instead

I tried that first, but in my testing, not every entity sends the death entity event packet (i.e Players), so this was the only way I could think of that would always work

EDIT Sep. 22: I may have figured out something, hold off on this

PR #5046, opened 2026-09-16
Add 26.3 support

andreypfau:
Registers protocol 777 and adds the 26.2 -> 26.3 translation.

### Packet ids

All three tables shift. Clientbound play gains `add_transient_block`, `post_effects` and `swing_animation`; clientbound configuration gains `post_effects`; serverbound play replaces `swing` with `punch`. The enums were generated from `GameProtocols`/`ConfigurationProtocols` rather than shifted by hand, and diffed against the 26.2 enums to confirm only those entries moved.

### Rewritten formats

- Relative entity movement packs on-ground and a step count into a properties var int, with the delta following as a `VecDelta`
- Entity position sync drops the delta movement and wraps the position in a `PositionPath`
- Level particles moves the particle to the front, splits the max speed per axis, widens the count to a var int and appends a randomization type
- Explode appends a play sound flag
- The previous game mode in login and respawn became a compact optional var int instead of a signed byte using -1 for none
- Advancements carry the tree layout, which moved out of the display data, so the coordinates are relocated rather than synthesized
- Sign editing and sign updates replace the front text boolean with a text slot enum, in both directions
- Swings moved out of `animate` into their own packet, and the remaining action ids were renumbered
- Player actions shift by the inserted `CHANGE_DESTROY_DIRECTION`, which is cancelled since it owes no acknowledgement
- Accept teleportation gained a position and rotation, discarded on the way to the server
- Bit sets are sent as byte arrays rather than long arrays, which affects light data in both chunk and light update packets, and partially filtered chat

Since 26.3 no longer sends a swing for item use, one is synthesized from `use_item` and `use_item_on` so other players still see the arm animation on a 26.2 server.

### Registry data

Loot conditions dispatch on the default key instead of `condition`, damage source tag predicates hold holder sets so their ids need the tag marker, trim materials replaced the asset group with a palette id, and block state provider types were renamed while block states moved to lowercase field names. `block_transformer` and `decorated_pot_pattern` are sent to the client from bundled data, since 26.2 servers do not know those registries and the client fails item initialization without the former.

### Item components

Twelve additions, `swing_animation` split into attack and interact animations, `map_color` dropped, and `pot_decorations` changed from a list of item ids into four optional item templates. Brick is the empty marker on the 26.2 side and is converted in both directions.

### Mappings

Generated with the existing `mapping-26.3.json` and diff files already present in the Mappings repository. Regenerating everything reproduced 55 of the committed files byte for byte, so only the three genuinely new or grown files are included here.

### Testing

Built and tested live: a real 26.3 client joined a vanilla 26.2 server through ViaProxy. Verified in game were world join, chunks and light, particles, TNT explosions, death and respawn, and item use. A test asserting every 26.3 data component id resolves to a serializer is included.

### Known gaps

- Loot conditions written as a bare list, which 26.2 accepted as an implicit `all_of`, are not converted. No vanilla data uses that form.
- `block_state_provider` is not sent; it has no vanilla entries.
- The 256 character command suggestion cap for clients without command block permission is not applied.
- `override_armor_assets` is dropped, so the darker trim palettes render with their base palette.
- Advancements without display data are positioned at the origin, since 26.2 servers never computed coordinates for them.


kennytv:
There's plenty missing here, we've done the work already and it'll become public on Friday

PR #5049, opened 2026-09-17
Regenerate 1.9 identifiers in the current mappings format

JobsonMarinho:
The `identifiers-1.9.nbt` file was added storing its sound mappings as separate `at`/`to` integer arrays without a `version` tag, matching the pre-v2 mapping layout. `MappingDataLoader` rejects it with the message "Invalid version: -1"; consequently, any protocol bridging versions 1.9 and 1.8—such as ViaRewind's `Protocol1_9To1_8`—finishes loading without any mapping data, as the `AbstractProtocol` class merely logs the failure.

This protocol continues translating packet fields but fails to rewrite packet IDs; as a result, a 1.8 client connected via BungeeCord receives the "join game" packet with the 1.9 version ID, and the proxy terminates the connection with the error "Unexpected packet received during server login process!".
javaw_WxlohZGCjn

The file needs to be regenerated in the current format, and a test must be added to fail if the format version is missing or if the file is unreadable by the loader. This ensures that data files generated in an obsolete format are detected at compile time, preventing a protocol from being silently disabled at runtime.


kennytv:
Thanks, I pushed the updated file myself since verifying binary patches is yuck and the tests are unnecessary, at least in that specific form

PR #5050, opened 2026-09-18
Fix armor HUD updates for server-driven equipment changes on 1.8

EthanKeating:
Players on newer clients can receive kit armor from a 1.8 server while their armor bar remains empty or stale until they manually re-equip it or reconnect. This was reported with a 26.2 client in Practice/UHC match setup, where the server equips enchanted armor through the Bukkit inventory API.

Via's Bukkit armor listener refreshes the emulated `generic.armor` attribute for player interactions and lifecycle events, but outgoing inventory updates do not refresh it. A world change can also calculate the attribute before the kit is equipped; same-world kit changes do not trigger that listener at all.

This change tracks the player's armor in the 1.8-to-1.9 protocol and sends the existing armor attribute/modifier when player inventory contents, armor slots, or self-equipment packets arrive. It also restores the tracked attribute after respawns. The Bukkit listener shares this tracker so manual equips continue using the same calculation. Bukkit inventory reads remain on the server thread, with tracker updates on the connection's event loop. Other containers and cursor updates do not affect the armor state.

Reproduction:

1. Connect a 26.2 client to a 1.8 Spigot server through ViaVersion.
2. Have a plugin equip an armor kit and update the player's inventory when entering a match, including a same-world teleport.
3. Observe that the armor items arrive but the armor bar can remain empty. Manually re-equipping the armor or reconnecting restores the bar.

Validation:

- `:viaversion:shadowJar :viaversion-common:test` passed; all 25 existing common-module tests passed.
- The patched jar started successfully on a local RivalsSpigot 1.8.8 Practice server behind Velocity. Both the proxy and backend answered 26.2 protocol status requests.
- No new automated regression tests were added; the existing suite does not directly exercise this armor scenario.


common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/storage/ArmorTracker.java
```diff
@@ -0,0 +1,62 @@
...
+
+public final class ArmorTracker implements StorableObject {
+
+    private static final UUID ARMOR_ATTRIBUTE = UUID.fromString("2AD3F246-FEE1-4E67-B886-69FD380BB150");
+
+    // Helmet through boots, matching player container slots 5 through 8.
+    private final int[] armorPoints = new int[4];
+
+    public void setArmor(int slot, int itemId) {
+        armorPoints[slot] = ArmorTypes1_8.findById(itemId).getArmorPoints();
```
kennytv: Track whether the values are dirty so you don't keep spamming attribute updates unnecessarily

bukkit-legacy/src/main/java/com/viaversion/viaversion/bukkit/listeners/v1_8to1_9/ArmorListener.java
```diff
@@ -43,31 +41,31 @@
 
 public class ArmorListener extends ViaBukkitListener {
 
-    private static final UUID ARMOR_ATTRIBUTE = UUID.fromString("2AD3F246-FEE1-4E67-B886-69FD380BB150");
-
     public ArmorListener(Plugin plugin) {
```
kennytv, line 41: Since you added a respawn packet handler, I assume the change world, join and respawn event listeners are no longer needed, though that should be tested

kennytv (changes requested):

bukkit-legacy/src/main/java/com/viaversion/viaversion/bukkit/listeners/v1_8to1_9/ArmorListener.java
EthanKeating: Tested, and you are correct. The changes have been made

common/src/main/java/com/viaversion/viaversion/protocols/v1_8to1_9/storage/ArmorTracker.java
EthanKeating: This is a great point, I made the changes

PR #5052, opened 2026-09-18
Restore 26.2 interpolation steps for 26.2->26.3

roccodev:
I think the time parameter for the interpolation in relative move packets should match 26.2, not 26.3. Specifically, 26.2 and below would use 3 ticks to interpolate all entity movement, while in 26.3 it is now based on the update timer.

This seems to match the server/older clients more closely. The difference is particularly noticeable when making small movements or when knocking back players, here's an example.

Before (small movements stutter, knockback is too 'fast'):

https://github.com/user-attachments/assets/3a94e565-1e3e-4a60-a89f-b1c20500a1ef

After:

https://github.com/user-attachments/assets/5ff23893-52b1-404d-84b2-3d4c72bbeafa

It seems to do well on all entities I've seen. Also I've added steps to position sync as well, it doesn't make a huge difference but it seems more correct


VeVeVeVel:
Have you verified if all entities, including projectiles, actually update every 3 ticks? To ensure full compatibility for this change, we'd likely need to check the tick update behavior across all entity types for versions 1.7 through 26.2.

roccodev:
I will in a moment, but it's worth noting that the client implementation seems to match:

In 26.2, `InterpolationHandler#interpolate` ([src](https://mcsrc.dev/2/26.2/net/minecraft/world/entity/InterpolationHandler#L77)) is called every tick for all entities with interpolation, and it just performs 1 lerp over the total steps, which is 3 for all entities with interpolation except displays (see same source).

In 26.3, there is a `Linear` type for `PositionPath`, but you can't configure the number of steps and it instead uses the entity's update interval, which desyncs it from 26.2. The ideal solution would be to use a `Stepped` with 3 steps offset by 1 tick each, and lerp the delta on the server side (could try that if people prefer it). However, at least for now, using 1 step with 3 ticks seems to work the same, because the game is still interpolating every tick within each step ([src](https://mcsrc.dev/2/26.3/net/minecraft/world/entity/SteppedInterpolationHandler#L178)).

Regarding old version compatibility, 1.8 is also using 3 ticks (and the videos were captured on 1.8 server), but I don't think it's too relevant anyway? 26.2 already doesn't really match 1.8 I don't think (at least to me, the interpolation looks a lot smoother on 26.2), so it's just a matter of making the difference in the 26.2->26.3 jump as minimal as possible.

Also upon further look, it seems that boats and displays still use a linear interpolator with 3 and 0 steps respectively in 26.3, the check in my PR isn't needed because it just ignores the tick offset for those entities, I'll simplify it.

roccodev:
For projectiles, etc. here are all the entities defining interpolation in 26.3:

interpolators

* `Entity`, `Shulker`, and `AbstractMinecart` with "experimental movement" use `InterpolationHandler.NO_OP`, which just teleports the entity on move.
* `LivingEntity` uses `SteppedInterpolationHandler`
* The rest use `LinearInterpolationHandler` with varying step count, but all of them ignore the tick offsets from the move packets. Notably, fishing hooks are the only projectiles that use interpolation from move packets.

So, living entities are the only ones concerned by this change, and all of them use 3 ticks in 26.2

kennytv (approved):

PR #5054, opened 2026-09-18
[26.2->26.3] Fix missing movement timestamp update and motion packet in ENTITY_POSITION_SYNC

Pantera07:
ENTITY_POSITION_SYNC was missing a couple of things: we weren't updating the movement timestamp, and the velocity wasn't actually being sent to the client. Because of this, the interpolation tick offset kept getting messed up.

Added the updateMovementTime call and made sure the SET_ENTITY_MOTION packet actually gets fired (ignoring duplicate zero motions).

Tested this in-game and entity movement is way smoother now. Knockback especially used to feel super jittery when hitting entities, but now it looks normal and fluid.


common/src/main/java/com/viaversion/viaversion/protocols/v26_2to26_3/rewriter/EntityPacketRewriter26_3.java
```diff
@@ -89,20 +99,42 @@ public void registerPackets() {
...
             wrapper.passthrough(Types.DOUBLE); // X
             wrapper.passthrough(Types.DOUBLE); // Y
             wrapper.passthrough(Types.DOUBLE); // Z
 
-            wrapper.read(Types.DOUBLE); // Delta x
-            wrapper.read(Types.DOUBLE); // Delta y
-            wrapper.read(Types.DOUBLE); // Delta z
+            final double deltaX = wrapper.read(Types.DOUBLE);
+            final double deltaY = wrapper.read(Types.DOUBLE);
+            final double deltaZ = wrapper.read(Types.DOUBLE);
```
kennytv, line 120: These were never actually used by the client, so I doubt this needs to be handled here now

kennytv (changes requested):

kennytv:
Closing in favor of https://github.com/ViaVersion/ViaVersion/pull/5052

PR #5058, opened 2026-09-19
Use shorten interpolation step ticks for entity types in 26.2->26.3

florianreuth:


roccodev:
Are you sure about this? When using stepped interpolation, the update interval value is only used to control the speed, which shouldn't affect much.

I tried 2 for players and small movements are stuttery as pointed out in #5052

florianreuth:
In my testing it makes movement feel smoother, and vanilla also has these step ticks in EntityTypes. Feel free to also do testing with this patch applied.

roccodev:
That's strange cause I'm having the opposite kinda haha

Here's a clip with 2 ticks, first part is the 26.3 pov then I try the same scenario on 26.2:
https://github.com/user-attachments/assets/a4312dd9-4af1-4d1c-b4e1-1f381e811128

I don't know if it is as noticeable for you too, but the knockback seems very slingshot-like with a very fast start and me almost getting sucked to the ground.

I think 3 ticks (this one here) is closer to 26.2, but maybe it's just preference?
https://github.com/user-attachments/assets/ca93df65-fbcf-4857-9e46-d0cd7aae9ffd

I guess it's understandable that with 3 ticks there's this sort of 'smoothing out' for faster movements (like the quick movement stutters in the PR I linked above)

Also worth noting I'm on a 1.8 server, did you test on a different environment?

florianreuth:
I tested on 26.2 server as a 1.8 server would mix tons of other changes into it as well, but I haven't taken a look at knockback yet. You should test on a 26.2 server though

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

kennytv:
It's also worth noting that the whole system isn't exactly fully functional on 26.3 either, so fully smooth movement likely isn't possible. The path of least issues currently seems to be the state of the main branch

roccodev:
Ah now that you mention it, I haven't tried native 26.3 multiplayer yet lol. Did Mojang really break it?

PR #5062, opened 2026-09-20
[26.2->26.3] Handle hand swing animations on block place and item drop

Pantera07:
26.3 clients don't play hand swing animations when dropping items or placing blocks on 26.2 servers.

These are handled through Bukkit listeners so we only trigger swings when the server actually allows the action:
- `BlockPlaceEvent`: swings main/off hand depending on the hand used.
- `PlayerDropItemEvent` & `InventoryClickEvent`: handles both hotbar drops and container border/slot drops.

Things like block interactions or throwing snowballs aren't covered here and can't be easily fixed this way. `PlayerInteractEvent` doesn't expose the server's `InteractionResult`, so there's no reliable way to verify if those actually succeeded without causing ghost swings.


roccodev:
Could this not use the block change acknowledged packet instead? Also `PlayerDropItemEvent` already fires for inventory drops, and you need to check that the player is the item's thrower because it also fires for /give

Pantera07:
Thanks for pointing that out!

I've removed `InventoryClickEvent` since `PlayerDropItemEvent` already covers inventory drops, preventing duplicate swing triggers.

Regarding `/give`, I verified on Paper 26.2 that dropping items when the inventory is full doesn't actually fire `PlayerDropItemEvent`. Additionally, 26.2 clients don't play a swing animation when receiving overflow items from `/give` anyway, so no extra checks should be needed here.

Pantera07:
I actually tried using `BLOCK_CHANGED_ACK`, but it doesn't work out.

The server still sends `BLOCK_CHANGED_ACK` on empty-hand clicks or failed placement attempts (like entity collision). In 26.3, the server only triggers the player's swing animation when the placement actually succeeds, so handling this via Bukkit's `BlockPlaceEvent` is necessary to prevent ghost swings.

